PR #11, opened 2019-09-30
Edit topology

dokshina:
Use `edit_topology` mutation to manage replicasets
Get rid of `leader` parameter for replicaset, specify instances list in failover priority order
Add `expelled` flag for instance to indicate if instance must me expelled


README.md
```diff
@@ -293,7 +299,9 @@ Environment=TARANTOOL_CONSOLE_SOCK=/var/run/tarantool/${app_name}.{instance_name
 To configure replicasets you need to specify replicaset parameters for each instance in replicaset:
 
 * `replicaset_alias` (`string`, optional) - replicaset alias, will be displayed in Web UI;
-* `leader` (`string`, required if `replicaset_alias` specified) - name of leader instance;
```
artur-barsegyan, line 306: At now does user have a way to mark an instance as a leader? It seems, not.

dokshina: The new way to mark instance as a leader is to specify it first in failover priority.

This configurations are equal:

```yaml
# Old 
replicaset_alias: replicaset-1
leader: replicaset-1-leader

# New
replicaset_alias: replicaset-1
failover_priority:
  - replicaset-1-leader
```

artur-barsegyan: It's not obvious for end-user. Your config (aka inventory) needs to be convenient, easy readable, and most of all **obvious**. 

You need to understand that we can deploy via `ansible-cartridge` only **master -> slave** replication configuration and we need to explicit specify, where is the leader.

artur-barsegyan: @vasiliy-t @knazarov @rosik we need your opinions.

README.md
```diff
@@ -162,7 +168,7 @@ This role tasks have special tags that allows to perform only secified actions.
 Tasks are running in this order:
 
 * `cartridge-instances` - install package, update instances config and restart instances;
-* `cartridge-replicasets` - configure replicasets;
+* `cartridge-replicasets` - configure replicasets, expell instances;
```
rosik: expel - one 'l'

README.md
rosik: > You need to understand that we can deploy via ansible-cartridge only master -> slave replication configuration

What do you mean, only master-slave? That's wrong. Cartridge as well as vshard always tries to set up full mesh master-master replication.

In general I like Liza's design. The only thing I would suggest is mentioning that the first server in failover_priority list is the leader. And make more cross-references describing all_rw, leadership, failover_priority and failover modes

artur-barsegyan: > And make more cross-references describing all_rw, leadership, failover_priority and failover modes

Seems for me it's too difficult and messy.

PR #41, opened 2019-11-19
Manage memtx_memory in runtime

dokshina:
Try to increase memtx_memory without instance restart (using
`cartridge_instance` module)
Add `cartridge_needs_restart` module to check if instance restart is
required
Add `restarted` variable to specify that instance must be restarted
Describe changes in README

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,6 +82,7 @@ all:
           config:
             advertise_uri: '172.19.0.2:3302'
             http_port: 8082
+          restarted: true  # force instance restart
 
     host2:
       vars:
@@ -152,6 +153,7 @@ Configuration format is described in detail in the
   indicates if the Tarantool repository should be enabled (for packages with
   open-source Tarantool dependency);
 * `config` (`dict`, required): instance configuration;
+* `restarted`(`boolean`, optional, default: `false`): indicates that instance must be restarted;
 * `replicaset_alias` (`string`, optional) - replicaset alias, will be displayed in Web UI;
 * `leader` (`string`, required if `replicaset_alias` specified) - name of leader instance;
 * `roles` (`list-of-strings`, required if `replicaset_alias` specified) - roles to be enabled on the replicaset.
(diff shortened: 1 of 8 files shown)


library/cartridge_instance.py
```diff
@@ -0,0 +1,76 @@
...
+        return type(box.cfg) ~= 'function' and box.cfg.memtx_memory or require('json').NULL
+    ''')
+
+    if new_memtx_memory <= current_memtx_memory:
+        return ModuleRes(success=True, changed=False)
+
+    # try to increase memtx_memory
+    increased = control_console.eval('''
+        local ok, err = pcall(function()
+            box.cfg {{ memtx_memory = {} }}
```
artur-barsegyan, line 52: Why you need doubled braces?

library/cartridge_instance.py
```diff
@@ -0,0 +1,76 @@
...
+        end)
+        if not ok then
+            if tostring(err):find("cannot decrease memory size at runtime") == nil then
```
artur-barsegyan, line 55: It's not a solution. You need to find another way. Try to google about `box.error` submodule: you need to get last box error.

README.md
```diff
@@ -152,6 +153,7 @@ Configuration format is described in detail in the
   indicates if the Tarantool repository should be enabled (for packages with
   open-source Tarantool dependency);
 * `config` (`dict`, required): instance configuration;
+* `restarted`(`boolean`, optional, default: `false`): indicates that instance must be restarted;
```
artur-barsegyan: hmm, _must_ or _should_?

library/cartridge_instance.py
```diff
@@ -0,0 +1,76 @@
...
+        new_memtx_memory = cartridge_defaults['memtx_memory']
+
+    # Check if instance is started
+    if not os.path.exists(control_sock):
+        return ModuleRes(success=True, changed=False)
+
+    control_console = get_control_console(control_sock)
+
+    # Get current memtx memory
+    current_memtx_memory = control_console.eval('''
```
artur-barsegyan, line 39: Where is the checks for `or require('json').NULL`?

library/cartridge_instance.py
artur-barsegyan, line 55: How it can be? You check the changes on decreasing some lines ago!

library/cartridge_instance.py
dokshina: Because it's a way to specify `{` and `}` symbols in format string.

README.md
dokshina: @lenkis help me, please!

I mean forced instance restarting.

library/cartridge_instance.py
dokshina: There are some cases when Tarantool can't do it in runtime anyway.

It depends on `box.slab.info().quota_size` value, but it's not documented anywhere, so I just check if Tarantool can do it in runtime or not.

PR #43, opened 2019-11-21
Test deb packages deployment

dokshina:


molecule/default/molecule.yml
```diff
@@ -20,11 +20,8 @@ platforms:
       - name: cartridge-network
     purge_networks: true
   - name: vm2
-    # XXX: wait for Tarantool fix
```
artur-barsegyan, line 23: What the fix?

artur-barsegyan (approved):
LGTM

molecule/default/molecule.yml
dokshina: This fix https://github.com/tarantool/tarantool/issues/4606.

PR #83, opened 2020-04-01
Unit tests

dokshina:


library/cartridge_manage_replicaset.py
```diff
@@ -25,13 +26,13 @@ def get_cluster_replicaset(control_console, name):
             end
 
             table.insert(res, {
-                uuid = r.uuid,
-                alias = r.alias,
-                status = r.status,
-                roles = r.roles,
-                all_rw = r.all_rw,
+                uuid = r.uuid or require('json').NULL,
```
olegrok: I think you could save json into local variable or use `box.NULL` without any require

olegrok (approved):

PR #86, opened 2020-04-08
Allow to downgrade package

### LINKED ISSUE #81: Make an ability to downgrade RPM
For possible rollback would be great to have an ability to downgrade RPM. There is an option for yum to allow_downgrade and it defaults to NO. Please make it possible to optionally switch it to YES.
https://docs.ansible.com/ansible/latest/modules/yum_module.html

dokshina:
Added `cartridge_allow_downgrade` option to specify that 
package downgrading is allowed

Closes #81

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ README.md and examples/getting-started-app/README.md
 to use the newest tag with new release
 -->
 
+### Added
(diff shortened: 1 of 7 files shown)


olegrok (approved):
I don't understand your patch. But I believe you.

PR #93, opened 2020-04-13
Set `any_errors_fatal: true` for package installation tasks

### LINKED ISSUE #79: Play proceed even when task "Copy RPM" fail
I have one machine and multiple tarantool instances in my `hosts.yml` file. Task "Copy RPM" triggered only on the first instance. If it fail, play proceed executing for other instances but not for one with failed task.

```
storage-1                  : ok=5    changed=0    unreachable=0    failed=1    skipped=1    rescued=0    ignored=0
storage-1-replica          : ok=29   changed=0    unreachable=0    failed=0    skipped=6    rescued=0    ignored=0
storage-2                  : ok=8    changed=0    unreachable=0    failed=0    skipped=4    rescued=0    ignored=0
storage-2-replica          : ok=8    changed=0    unreachable=0    failed=0    skipped=4    rescued=0    ignored=0
```

Adding option `any_errors_fatal` to role task helped.

Consider to make this behaviour as default.

dokshina:
Package installation tasks are running once for all instances on one physical machine.
It mean that when this task is failed, all play should fail immediately.

Closes #79

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ README.md and examples/getting-started-app/README.md
 to use the newest tag with new release
 -->
 
+### Changed
+
+- `any_errors_fatal: true` is set for package installation tasks
+
 ## [1.2.0] - 2020-04-08
 
 ### Fixed
--- a/README.md
+++ b/README.md
@@ -56,6 +56,7 @@ To deploy an application and set up this topology:
   hosts: all
   become: true
   become_user: root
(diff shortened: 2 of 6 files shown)


RunsFor:
Lgtm, but I'm sure there should be a more cleaner way to get around of these. We'll see.

Also, I would add some commentary text in a commit for those who won't search for GitHub issues for the motivation behind this.

PR #96, opened 2020-04-27
Setting stateful failover parameters

### LINKED ISSUE #95: Support stateful failover
This is the last part of https://github.com/tarantool/cartridge/issues/714

Remind, that stateboard systemd unit (`/etc/systemd/system/<appname>-stateboard.service`) is delivered in application package **only** if application contains `stateboard.init.lua` in it's root directory.

It reads configuration from `<appname>-stateboard` section in `/etc/tarantool/conf.d` files.

dokshina:
This is the first part of #95.

New failover parameters are support:
* `stateful` and `eventual` modes
* `stateboard` state provider options for `stateful` failover


library/cartridge_manage_failover.py
```diff
@@ -39,6 +49,80 @@ def manage_failover(params):
...
+            if lua_stateboard_params:
+                lua_params.append('tarantool_params = {{ {} }}'.format(', '.join(lua_stateboard_params)))
+
+    res = control_console.eval('''
+        local ok, err = require('cartridge').failover_set_params({{
+            {}
+        }})
+
+        return {{
+            ok = ok or box.NULL,
```
olegrok: Just notice that `false or box.NULL -> box.NULL`

library/cartridge_manage_failover.py
```diff
@@ -39,6 +49,80 @@ def manage_failover(params):
...
+    ]
+
+    if mode == 'stateful':
+        state_provider = failover_params.get('state_provider')
+        if state_provider == 'stateboard':
+            lua_params.append('state_provider = "tarantool"')
+
+            stateboard_params = failover_params.get('stateboard_params')
+
+            lua_stateboard_params = []
```
olegrok, line 70: It's just strange way to concat an arguments to string and eval then.

If you use tarantool-python connector you can form argument that you need in Python and then:
```py
arg = {"key": value}
conn.eval('''
    local arg = ...
    func(arg)
''', arg)
```

dokshina: I don't use `tarantool-python` connector, but I'm planning to improve passing arguments for Lua API calls.

olegrok (approved):

Issue #102, opened 2020-05-15
Support custom instance socket path


vanyarock01:
## Summary

Add the ability to specify a custom path to the socket for every host.

## Motivation

The directory `/var/run/tarantool/` can not always be reached. This makes it impossible to use a role in such cases.

## Design Detail

Add optional string parameter `control_sock_path` to instance hostvars and change the function to get the socket.

```python
def get_instance_control_sock(instance_vars, stateboard=False):
    if instance_vars.get('control_sock_path'):
        return instance_vars['control_sock_path']

    return '/var/run/tarantool/{}.control'.format(
        get_instance_fullname(
            instance_vars['cartridge_app_name'],
            instance_vars['inventory_hostname'],
            stateboard))
```

Filter call will look like this:
```yaml
join_instance_sock: '{{ hostvars[join_host] | get_instance_control_sock }}'
```
## Drawbacks

When the ansible-cartridge role starts the instance, the default path `/var/run/tarantool` will be passed to it through from systemd unit. That is, specifying a custom path will not create a file socket, it must already exist.
Therefore, this feature is only suitable for managing an already created instances.

dokshina:
Since #199 it's possible to pass `cartridge_run_dir` variable.

PR #107, opened 2020-07-09
Add supporting of custom instance socket path

vanyarock01:
Related issue  #102

### DIFF
--- a/filter_plugins/filter_hosts.py
+++ b/filter_plugins/filter_hosts.py
@@ -10,11 +10,11 @@ def is_stateboard(instance_vars):
     return instance_vars.get('stateboard') is True
 
 
-def get_instance_fullname(app_name, inventory_hostname, stateboard):
-    if stateboard:
-        return '{}-stateboard'.format(app_name)
+def get_instance_fullname(instance_vars):
+    if instance_vars.get('stateboard'):
+        return '{}-stateboard'.format(instance_vars['app_name'])
 
-    return '{}.{}'.format(app_name, inventory_hostname)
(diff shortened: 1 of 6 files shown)


unit/test_filter_hosts.py
```diff
@@ -0,0 +1,31 @@
...
+
+    def test_compose_instance_control_sock(self):
+        instance_hostvars = {
+            'cartridge_app_name': 'app_name',
+            'inventory_hostname': 'host_name'
+        }
+        self.assertEqual(
+            self.filter_table['get_instance_control_sock'](instance_hostvars),
+            '/var/run/tarantool/{app_name}.{host_name}.control'.format(
+                instance_hostvars['cartridge_app_name'],
```
dokshina: Looks like keyword args are missed here.

filter_plugins/filter_hosts.py
```diff
@@ -43,9 +43,15 @@ def get_one_not_expelled_instance(hostvars, play_hosts):
     raise Exception('At least one play host should be non-expelled and not stateboard')
 
 
-def get_instance_control_sock(app_name, inventory_hostname, stateboard=False):
-    instance_fullname = get_instance_fullname(app_name, inventory_hostname, stateboard)
-    return '/var/run/tarantool/{}.control'.format(instance_fullname)
+def get_instance_control_sock(instance_vars):
```
dokshina, line 46: I suggest changing other filter functions signatures to receive `instance_vars` arguments.

dokshina (commented):
Thanks a lot for your patch!

Please, fix review comments and add a documentation for the new variable.

dokshina:
Since #199 it's possible to pass `cartridge_run_dir` variable.

PR #116, opened 2020-09-29
Update Cartridge to 2.3.0 in tests

### LINKED ISSUE #115: [1pt] Fix edit user idempotence check for cartridge 2.3.0
In a new versions of cartridge, user has `version` field that is changed on every user updte. This field should be ignoring on comparing new user with the old one.

dokshina:
Closes #115

### DIFF
--- a/library/cartridge_auth.py
+++ b/library/cartridge_auth.py
@@ -157,6 +157,27 @@ def user_is_deleted(user):
     return 'deleted' in user and user['deleted'] is True
 
 
+def users_are_equal(user1, user2):
+    for key in user1:
+        if key == 'version':
+            continue
+
+        if key not in user2:
+            return False
+
+        if user1[key] != user2[key]:
+            return False
+
+    for key in user2:
+        if key == 'version':
+            continue
+
+        if key not in user1:
+            return False
+
+    return True
+
+
 def manage_auth(params):
     auth_params = params['auth']
     control_console = get_control_console(params['control_sock'])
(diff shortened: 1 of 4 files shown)


library/cartridge_auth.py
```diff
@@ -157,6 +157,27 @@ def user_is_deleted(user):
     return 'deleted' in user and user['deleted'] is True
 
 
+def users_are_equal(user1, user2):
+    for key in user1:
+        if key == 'version':
```
olegrok, line 162: Why not smth similar to
```py
user1_cpy = user1.copy()
user2_cpy = user2.copy()

# Maybe additional check needed here
del user1_cpy['version']
del user2_cpy['version']

return user1_cpy == user2_cpy
```

unit/test_auth.py
```diff
@@ -249,6 +249,65 @@ def test_edit_user(self):
...
+
+        USER = {
+            'username': 'dokshina',
+            'email': 'dokshina@tarantool.love',
+            'fullname': 'Elizaveta Dokshina',
+        }
+
+        new_params = {
+            'password': 'new-password',
+            'email': 'dokshina@tarantool.hate',
```
olegrok:
```suggestion
            'email': 'dokshina@tarantool.love_a_bit_more',
```

library/cartridge_auth.py
dokshina: What's the difference?

olegrok: My suggestion is shorter

olegrok (approved):

PR #123, opened 2020-11-02
CI: Move to GitHub Actions

dokshina:


olegrok (approved):

.github/workflows/molecule-tests.yml
```diff
@@ -0,0 +1,124 @@
...
+  pull_request:
+
+env:
+  CARTRIDGE_CLI_VERSION: '2.4.0'
+
+jobs:
+  molecule-tests-ce:
+    if: |
+      github.event_name == 'push' ||
+      github.event_name == 'pull_request' && github.event.pull_request.head.repo.owner.login != 'tarantool'
```
RunsFor, line 15: Why is this check needed

.github/workflows/molecule-tests.yml
```diff
@@ -0,0 +1,124 @@
...
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        tarantool-version: ["1.10", "2.5"]
+      fail-fast: false
+    steps:
+      - uses: actions/checkout@master
```
RunsFor: It will be more stable to bind to a specific version. I suppose, `v2` would be enough

.github/workflows/molecule-tests.yml
```diff
@@ -0,0 +1,124 @@
...
+      - name: Create test packages
+        if: steps.cache-packages.outputs.cache-hit != 'true'
+        working-directory: '${{ github.repository }}'
+        run: |
+          sudo apt -y update
+          sudo apt -y install git gcc make cmake unzip
+          git config --global user.email "test@tarantool.io" \
+            && git config --global user.name "Tar Antool"
+
+          curl -L https://tarantool.io/JRceXH/release/${{ matrix.tarantool-version }}/installer.sh | bash
```
RunsFor: Should work without `JRceXH`

molecule/default/molecule.yml
```diff
@@ -18,7 +18,6 @@ platforms:
       - 8081-8090:8081-8090/tcp
     networks:
       - name: cartridge-network
-    purge_networks: true
```
RunsFor, line 21: Is it still necessary after manually installing molecule?

RunsFor (commented):
Left some minor comments and questions

.github/workflows/molecule-tests.yml
dokshina: To run tests for only one of PR or push, not both.

.github/workflows/molecule-tests.yml
dokshina: This command is copied from [tarantool.io download](https://www.tarantool.io/en/download/os-installation/ubuntu/) instructions. I believe that it should be stable.

molecule/default/molecule.yml
dokshina: What do you mean? 
I think that this option was specified for tests on GitLab CI, now everything works fine without purging networks.

.github/workflows/molecule-tests.yml
RunsFor: This should solve it:


```yaml
on:
   push:
      branches: ['master']
   pull_request:
```

molecule/default/molecule.yml
RunsFor: `purge_networks` seems useful option which helps isolating a container. But I don't have any strong argument against returning it back.

.github/workflows/molecule-tests.yml
dokshina: Removed `JRceXH `.

.github/workflows/molecule-tests.yml
dokshina: We want to run workflow even if branch was pushed without PR.

PR #125, opened 2020-11-05
Manage dynamic box.cfg params without restart

### LINKED ISSUE #121: [4pt] Additional tarantool runtime parameters
Need more flexible tarantool configuration. Now only the memtx_memory parameter can be updated at runtime. I would like to update the rest of the parameters (e.g. box.cfg.readahead, box.cfg.net_msg_max...)

dokshina:
Closes #121

### DIFF
--- a/README.md
+++ b/README.md
@@ -324,6 +324,19 @@ Environment=TARANTOOL_PID_FILE=/var/run/tarantool/${app_name}.{instance_name}.pi
 Environment=TARANTOOL_CONSOLE_SOCK=/var/run/tarantool/${app_name}.{instance_name}.control
 ```
 
+#### Managing dymanic `box.cfg` parameters without restart
+
+Some `box.cfg` options are dynamic, it means that it can be changed without
+instance restarting.
+See [parameters](https://www.tarantool.io/en/doc/latest/reference/configuration/#configuration-parameters)
+with "Dynamic: yes".
+
+Role changes this parameters without restarting the instance.
+If other parameters are changed, instance is restarted anyway.
+
+**Note**, that if `restarted` flag is set, instance will be restarted anyway without changing dynamic parameters in runtime.
+You can use this flag to force instance restarting.
+
 #### Increasing memtx_memory in runtime
 
 If you specified in `config.memtx_memory` value that increases current `memtx_memory`, this role will try to increase this value in runtime.
--- a/library/cartridge_instance.py
+++ b/library/cartridge_instance.py
@@ -3,6 +3,7 @@
 from ansible.module_utils.basic import AnsibleModule
 from ansible.module_utils.helpers import ModuleRes, CartridgeException, cartridge_errcodes
 from ansible.module_utils.helpers import get_control_console
+from ansible.module_utils.helpers import dynamic_box_cfg_params, memory_size_box_cfg_params
 
 import os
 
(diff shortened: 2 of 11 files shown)


library/cartridge_instance.py
```diff
@@ -14,21 +15,115 @@
...
+
+def box_cfg_was_called(control_console):
+    return control_console.eval('''
+        return type(box.cfg) ~= 'function'
+    ''')
+
+
+def get_box_cfg(control_console):
+    return control_console.eval('''
+        return box.cfg or box.NULL
```
olegrok: Would you like to check that box.cfg is a table here?

library/cartridge_instance.py
```diff
@@ -14,21 +15,115 @@
...
+        new_memory_size = config[param_name]
+    else:
+        new_memory_size = cartridge_defaults[param_name]
+
+    if new_memory_size is None:
+        return False
+
+    # Get current memory size
+    current_memory_size = control_console.eval('''
+        return type(box.cfg) ~= 'function' and box.cfg.{} or box.NULL
```
olegrok, line 64: Are you sure `box.cfg{}` implicitly should be called here instead of return an error?

olegrok: So, it seems a typo here `box.cfg.{}` -> `box.cfg{}`

library/cartridge_instance.py
```diff
@@ -14,21 +15,115 @@
...
+    if new_memory_size <= current_memory_size:
+        return False
+
+    # try to increase memory size
+    increased = control_console.eval('''
+        local ok, err = pcall(function()
+            box.cfg {{ {param_name} = {new_memory_size} }}
+        end)
+        if not ok then
+            if tostring(err):find("cannot decrease memory size at runtime") == nil then
```
olegrok, line 84: Seems you lost an error in case when `tostring(err):find(...) ~= nil`

README.md
```diff
@@ -324,6 +324,19 @@ Environment=TARANTOOL_PID_FILE=/var/run/tarantool/${app_name}.{instance_name}.pi
 Environment=TARANTOOL_CONSOLE_SOCK=/var/run/tarantool/${app_name}.{instance_name}.control
 ```
 
+#### Managing dymanic `box.cfg` parameters without restart
+
+Some `box.cfg` options are dynamic, it mean that it can be changed without
```
olegrok: it mean**s**

README.md
```diff
@@ -324,6 +324,19 @@ Environment=TARANTOOL_PID_FILE=/var/run/tarantool/${app_name}.{instance_name}.pi
...
+instance restarting.
+See [parameters](https://www.tarantool.io/en/doc/latest/reference/configuration/#configuration-parameters)
+with "Dynamic: yes".
+
+Role changes this parameters without restarting the instance.
+If other parameters were changed, instance is restarted anyway.
```
olegrok: `were` -> `are`.

Otherwise you need to correct your mixed conditional (https://engblog.ru/mixed-conditionals)

library/cartridge_instance.py
```diff
@@ -14,21 +15,115 @@
...
+    for k, v in d.items():
+        value_str = None
+        if type(v) == str:
+            value_str = "'%s'" % v
+        elif type(v) == float or type(v) == int:
+            value_str = "%s" % v
+        elif type(v) == bool:
+            value_str = 'true' if v else 'false'
+        else:
+            raise CartridgeException(
```
olegrok, line 30: It doesn't looks like "CartridgeException". E.g. "SerializeException"... but it's not related to cartridge

library/cartridge_instance.py
```diff
@@ -14,21 +15,115 @@
...
+    ''')
+
+
+def get_box_cfg(control_console):
+    return control_console.eval('''
+        return box.cfg or box.NULL
+    ''')
+
+
+def change_memory_size(param_name, cartridge_defaults, config, control_console):
```
olegrok, line 60: I'm not sure you need handle memory change separately. It should be handled as others dynamic options. Yes, it can't be decreased but you need just rethrow box error

library/cartridge_instance.py
dokshina: It's called `CartridgeException` for historical reasons, all such exceptions aren't related to cartridge.

library/cartridge_instance.py
dokshina: The main difference between memory changing and other dynamic params is that memory size can be only increased in runtime (but it still can be decreased by instance restart), but all other dynamic params can be changed in runtime. So

* if memory changing fails, there are two cases:
  * "cannot be decreased error" -> it's impossible to change memory in runtime, need to restart
  * all other errors -> fail, something went wrong
* if any other parameter changing fails there is only one case:
  * fail, something went wrong

library/cartridge_instance.py
dokshina: No, in this case `false` is returned. It mean that memory wasn't changed in runtime, so instance should be restarted to apply new configuration.

library/cartridge_instance.py
dokshina: Yes, I agree. 

I'm trying to get rid of this check in every eval, see `box_cfg_was_called` function.

library/cartridge_instance.py
```diff
@@ -14,21 +15,115 @@
...
+    new_memory_size = None
+    if param_name in config:
+        new_memory_size = config[param_name]
```
olegrok: `new_ new_memory_size = config.get(param_name, cartridge_defaults[param_name])`

library/cartridge_instance.py
```diff
@@ -14,21 +15,115 @@
...
+def change_dynamic_params(cartridge_defaults, config, control_console):
+    params = dict()
+
+    # Get current values
+    current_box_cfg = get_box_cfg(control_console)
+    if current_box_cfg is None:
+        return False
+
+    for param_name, param_value in config.items():
+        if param_name in dynamic_box_cfg_params and param_name not in memory_size_box_cfg_params:
```
olegrok: I think you need some one-line helpers here to improve readability: `is_dynamic_parameter`/`is_memory_management_parameter` to eliminate strange chains of conditions

olegrok (approved):

PR #128, opened 2020-11-10
Support stateboard deploy with tgz

opomuc:


filter_plugins/filter_hosts.py
```diff
@@ -78,6 +78,14 @@ def get_instance_systemd_service(app_name, inventory_hostname, stateboard):
     return '{}@{}'.format(app_name, inventory_hostname)
 
 
+def is_stateboard_on_machine(hosts, hostvars, inventory_hostname):
+    machine_id = get_machine_identifier(hostvars[inventory_hostname])
+    for i in hosts:
+        if get_machine_identifier(hostvars[i]) == machine_id and 'stateboard' in hostvars[i] and hostvars[i]['stateboard'] == True:
```
dokshina: Maybe `hostvars[i]['stateboard'] is True`?

tasks/install_targz.yml
```diff
@@ -15,12 +15,14 @@
...
+    owner: "{{ instance_user }}"
+    group: "{{ instance_group }}"
   loop:
     - {path: "{{ bin_dir }}{{ cartridge_app_name }}"}
     - {path: "{{ data_dir }}"}
     - {path: "{{ conf_dir }}"}
-    - {path: "{{ instance_run_path }}", mode: 777}
-    - {path: "{{ instance_systemd_path }}"}
+    - {path: "{{ instance_run_path }}"}
+    - {path: "{{ instance_work_dir }}"}
```
dokshina: Should it be ran for stateboard?
I remember that functions like a `get_instance_work_dir` accept `stateboard` flag and compute stateboard-specific directories, so there is no reason to run `Create stateboard folders` task.

filter_plugins/filter_hosts.py
```diff
@@ -78,6 +78,14 @@ def get_instance_systemd_service(app_name, inventory_hostname, stateboard):
     return '{}@{}'.format(app_name, inventory_hostname)
 
 
+def is_stateboard_on_machine(hosts, hostvars, inventory_hostname):
```
dokshina: I'm not sure that this check is needed. This role interprets each instance as Ansible host, so you don't need to run stateboard-specific tasks for each instance(Ansible host) placed on the same machine(physical machine, not Ansible host) as statebord. All you need is `stateboard` flag that indicated that tasks are running for stateboard instance(Ansible host).

tasks/install_targz.yml
```diff
@@ -50,15 +52,41 @@
...
+        state: directory
+        mode: "{{ item.mode | default('0755') }}"
+        owner: "{{ instance_user }}"
+        group: "{{ instance_group }}"
+      loop:
+        - {path: "{{ stateboard_work_dir }}"}
+        - {path: "{{ stateboard_run_dir }}"}
+      any_errors_fatal: true
+
+    - name: Create a stateboard service file
```
dokshina: I think that we can always create this file, just like RPM and DEB packages do.
For example, you unpacked TGZ and created a cluster without specifying stateboard instance and then you want to add stateboard instance without reinstalling TGZ. Since you didn't specified that you need stateboard on first TGZ installing, you don't have stateboard unit file.

tasks/install_targz.yml
```diff
@@ -50,15 +52,41 @@
...
+        src: stateboard.service.j2
+        dest: "{{ instance_systemd_path }}{{ stateboard_name }}.service"
+      any_errors_fatal: true
+      notify:
+        - reload-systemd-daemon
+  tags: always
+  when: stateboard_on_machine == true
+
+-  name: trigger reload-systemd-daemon
+   command: /bin/true
```
dokshina: It can be triggered only if previous tasks changed something.
Triggers are flushed once after package installing, so it's OK to add notify for reach task that changes files (in fact, we need to reload systemd daemon only if unit files were changed).

filter_plugins/filter_hosts.py
opomuc: `install_targz.yml` may not run on a stateboard instance (according to `get_one_not_expelled_instance_for_machine` filter). This check is needed to know when are we dealing with host, that is on the same machine as stateboard.

I do not know though, what to do when stateboard is on a separate machine... but that scenario has never been supported, so we can open an issue, I guess

tasks/install_targz.yml
opomuc: handlers are only triggered when a change happens. For some reason, I came across cases, when changing or creating a service file did not set "change" state. I try to reproduce it now...

opomuc: Found it: changing service file does not trigger instance restart. This may be required when switching between `ansible-cartridge` version.

But I see what you were trying to do, so I will remove this functionality. And create an issue

filter_plugins/filter_hosts.py
dokshina: You don't need to run `Configure stateboard` tasks. `stateboard_work_dir` and `stateboard_work_dir` are created on systemd service start (see unit file). I suggest you to place stateboard unit file always when TGZ is deployed (see comment below).

BTW, `get_one_not_expelled_instance_for_machine` works for stateboard on separate machine. I've just deployed this configuration  (see `molecule/default/hosts.yml`, I've removed all other instances and replicasets configurations):

```yaml
...
        host1:
          vars:
            ansible_host: vm1
            cartridge_package_path: ./packages/myapp-1.0.0-0.rpm

          hosts:
            my-stateboard:

        host2:
          vars:
            ansible_host: vm2
            cartridge_package_path: ./packages/myapp-1.0.0-0.deb

          hosts:
            core-1:
```

dokshina:
Thanks a lot for making this great job! 

We've used it in #199 - this PR introduced deploying TGZ as well as deploying stateboard with TGZ.

opomuc:
My pleasure!

PR #138, opened 2020-12-08
Extend readiness check with Cartridge state check

opomuc:
This is my proposal. Rework it if needed

### DIFF
--- a/library/cartridge_check_instance_started.py
+++ b/library/cartridge_check_instance_started.py
@@ -8,10 +8,12 @@ from ansible.module_utils.helpers import get_control_console
 argument_spec = {
     'control_sock': {'required': True, 'type': 'str'},
     'stateboard': {'required': True, 'type': 'bool'},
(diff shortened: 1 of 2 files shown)


library/cartridge_check_instance_started.py
```diff
@@ -8,21 +8,40 @@
...
 def check_instance_started(params):
+    waiting_for_state = params['waiting_for_state']
     try:
         control_console = get_control_console(params['control_sock'])
         ok = True
         if not params['stateboard']:
             ok = control_console.eval('''
                 return require('membership').myself().status == 'alive'
             ''')
+            if ok is True:
```
RunsFor, line 24: Redundant `if`, I think

opomuc: Probably, yes. In that case return `ok == False` shouldn't be `Instance is not running` as well. Is `unhealthy` a membership status as well? in that case it will fail but instance IS running, not healthy though

hackallcode:
Resolved by #176

PR #162, opened 2021-02-01
[2pt] Fix Molecule tests, CI and Shell scripts

hackallcode:
- Fix Shell scripts to run on OS X
- Update Python dependencies
- Fix tests and CI scripts seeing new Molecule

### DIFF
--- a/module_utils/helpers.py
+++ b/module_utils/helpers.py
@@ -30,7 +30,7 @@ cartridge_errcodes = CartridgeErrorCodes()
 
 
 # parameters of instance configuration that can be changed dynamically
-dynamic_box_cfg_params = set([
+dynamic_box_cfg_params = {
     'memtx_memory',
     'vinyl_memory',
     'custom_proc_title',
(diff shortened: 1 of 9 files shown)


hackallcode:
The CI step "Create Test Packages" and, accordingly, the `create-packages.sh` script were tested on the first run of current commit tests. On the second run of the tests, the CI step "Cache test packages" was tested.

module_utils/helpers.py
```diff
@@ -61,12 +61,12 @@ def __init__(self):
...
     'feedback_interval',
-])
+}
 
-memory_size_box_cfg_params = set([
+memory_size_box_cfg_params = {
     'memtx_memory',
     'vinyl_memory',
-])
+}
```
dokshina, line 69: Please, explain this change.

hackallcode: `set(['a', 'b', 'c'])` first creates a list, and then converts it to a set. `{'a', 'b', 'c'}` immediately creates a set. The IDE also generates a warning when you use the `set(['a', 'b', 'c'])` syntax.

But if you think that the previous version was more readable, then I will return it.

dokshina: I'm not sure that this change should be performed in this patch. You can perform it in separated commit ant merge this PR without squash.

hackallcode: Okey, I made two commits

dokshina (approved):
Thanks a lot!

PR #163, opened 2021-02-02
Use newest cartridge for tests

dokshina:


create-packages.sh
```diff
@@ -21,7 +21,10 @@ version=1.0.0-0
 pack_flags=''
 if [[ $(tarantool -V) == *"Target: Darwin"* ]]; then
     pack_flags+=' --use-docker'
-    if [ -z "${TARANTOOL_SDK_PATH}" ]; then
+fi
+
+if [[ $(tarantool -V) == *"Tarantool Enterprise"* ]]; then
```
hackallcode: Why if you have Tarantool Enterprise, you should set TARANTOOL_SDK_PATH?

dokshina: I've fixed that. If you run this script on OS X and use TE, your package will contain your current SDK (which is sdk-macos and doesn't work on Linux) or the one specified in TARANTOOL_SDK_PATH (you should specify here Linux SDK).

hackallcode: Okay, now it's better. But why you don't want to use "and" operation instead 2 "if"?

dokshina: Fixed.

hackallcode (approved):

PR #165, opened 2021-02-02
Call `edit_topology` only once

### LINKED ISSUE #143: [4pt] Call `edit_topology` only once
`edit_topology` request has been designed to bundle all replicaset change in one.

For example:
We faced an issue, when adding new storages to an existing cluster. When changing `replicaset_weight` from 0 to 1 on multiple replicasets `ansible-cartridge` started applying changes one-by-one. This leads to N rebalancing processes instead of one. Also, it is guranteed to fail since during rebalancing instances are in ConfiguringRoles stage, for some reason...

dokshina:
Closes #143


filter_plugins/filter_hosts.py
```diff
@@ -43,6 +43,18 @@ def get_one_not_expelled_instance(hostvars, play_hosts):
     raise Exception('At least one play host should be non-expelled and not stateboard')
 
 
+def get_join_host(control_host, hostvars, play_hosts):
+    if control_host:
+        return control_host
+
+    for i in play_hosts:
```
hackallcode: Is `i` a numeric counter here? If not, it is better to use a different variable name

filter_plugins/filter_hosts.py
```diff
@@ -43,6 +43,18 @@ def get_one_not_expelled_instance(hostvars, play_hosts):
     raise Exception('At least one play host should be non-expelled and not stateboard')
 
 
+def get_join_host(control_host, hostvars, play_hosts):
```
rosik: Let's rename it to `control_host`

rosik (approved):

library/cartridge_edit_topology.py
```diff
@@ -0,0 +1,408 @@
...
+
+def get_cluster_instances(control_console):
+    instances, _ = control_console.eval_res_err(get_instances_func_body)
+
+    return instances
+
+
+def get_configured_replicasets(hostvars, play_hosts):
+    replicasets = {}
+    for i, instance_vars in hostvars.items():
```
hackallcode: Is `i` a numeric counter here too?

unit/test_edit_topology.py
```diff
@@ -0,0 +1,589 @@
...
+            'weight': 1,
+            'vshard_group': 'group-1'
+        }
+
+        add_replicaset(
+            self.instance, rpl1_vars['replicaset_alias'], rpl1_vars['roles'],
+            ['r1-leader', 'r1-replica', 'r1-replica-2'],
+            all_rw=rpl1_vars['all_rw'], weight=rpl1_vars['weight'],
+            vshard_group=rpl1_vars['vshard_group']
+        )
```
hackallcode: Let's use keys for all args, it's more readble:
```suggestion
        add_replicaset(
            instance=self.instance,
            alias=rpl1_vars['replicaset_alias'], 
            roles=rpl1_vars['roles'],
            servers=['r1-leader', 'r1-replica', 'r1-replica-2'],
            all_rw=rpl1_vars['all_rw'],
            weight=rpl1_vars['weight'],
            vshard_group=rpl1_vars['vshard_group']
        )
```

filter_plugins/filter_hosts.py
dokshina: Yes, you are right. I'll use `instance_name` in this diff. Other such cases I would like to fix in separated patch.

tasks/main.yml
```diff
@@ -100,30 +100,20 @@
...
-  include_tasks: manage_replicaset.yml
-  vars:
-    replicaset: '{{ item }}'
-    join_host: '{{ replicasets.meta.join_host }}'
-  loop: '{{ replicasets.meta.replicasets }}'
+- name: Edit topology
+  cartridge_edit_topology:
+    hostvars: '{{ hostvars }}'
+    play_hosts: '{{ play_hosts }}'
+    healthy_timeout: '{{ replicaset_healthy_timeout }}'
```
hackallcode: Please, remove `healthy_timeout` param from `cartridge_edit_topology`

hackallcode (approved):

PR #166, opened 2021-02-02
Optimize Molecule sequences and fix Ansible check step

hackallcode:
Before the patch:
- Molecule tried to execute unused steps and issued warnings when it could not find them;
- Molecule `check` step failed.

This patch includes:
- Custom Molecule sequences;
- Ansible scripts are fixed so the Molecule `check` step passes without errors;
- Molecule `check` step added to CI.

### DIFF
--- a/tasks/install_deb.yml
+++ b/tasks/install_deb.yml
@@ -16,6 +16,7 @@
     warn: false
   register: deb_info
   changed_when: false
+  check_mode: false
   any_errors_fatal: true
 
 - name: Get package name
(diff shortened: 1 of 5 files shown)


tasks/install_deb.yml
```diff
@@ -16,6 +16,7 @@
     warn: false
   register: deb_info
   changed_when: false
+  check_mode: no
```
dokshina: Please, use `check_mode: false`

hackallcode: However, the documentation says to use the values yes or no:
https://docs.ansible.com/ansible/latest/user_guide/playbooks_checkmode.html#enforcing-or-preventing-check-mode-on-tasks

dokshina (approved):

PR #167, opened 2021-02-03
Use Cartridge tags for all tasks in playbook

### LINKED ISSUE #156: [2pt] Impossible to skip role using tags
I have several tasks in my playbook. One of them is importing the anisble-cartridge role.
I made a tag for this task like this:
```yaml
- name: Deploy Tarantool Cartridge app
  hosts: cartridge_group
  become: true
  any_errors_fatal: true
  tags:
    - cartridge
  tasks:
    - name: Import Tarantool Cartridge role
       import_role:
         name: tarantool.cartridge
```
After that, I start the playbook with the `-t some-tag` parameter. I expected the role installation to be skipped completely, but in the end I get error like this:
```
fatal: [host_name]: FAILED! => {"msg": "The conditional check 'ansible_os_family not in [\"RedHat\", \"Debian\"]' failed. The error was: error while evaluating conditional (ansible_os_family not in [\"RedHat\", \"Debian\"]): 'ansible_os_family' is undefined\n\nThe error appears to be in '.ansible/roles/tarantool.cartridge/tasks/main.yml': line 2, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n---\n- name: Check distribution\n  ^ here\n"}
```

I think this is because you are using `tags: always` instead list of tags.

hackallcode:
Closes #156

Before the patch, if you run the playbook with the `-t some-tag` option, some tasks (that have the `always` tag) will run. This made it difficult to skip the role installation.

Now role installation will be completely skipped if you specify a tag other than the tags for this role.

### DIFF
--- a/unit/instance.py
+++ b/unit/instance.py
@@ -38,6 +38,19 @@ class Instance:
         self.process = None
         self.sock = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
 
+        self.__original_exists = os.path.exists
+        self.__original_getmtime = os.path.getmtime
+
+        os.path.exists = OsPathExistsMock()
+        os.path.getmtime = OsPathGetMtimeMock()
+
+    def __del__(self):
+        os.path.exists = self.__original_exists
+        os.path.getmtime = self.__original_getmtime
+
+        if os.path.exists(self.console_sock):
+            os.remove(self.console_sock)
+
     @tenacity.retry(wait=tenacity.wait_fixed(0.1),
                     stop=tenacity.stop_after_delay(5))
     def _connect(self):
@@ -54,9 +67,6 @@ class Instance:
             self.process = Popen(command, env=env, stdout=FNULL, stderr=FNULL)
 
         self._connect()
-
-        os.path.exists = OsPathExistsMock()
-        os.path.getmtime = OsPathGetMtimeMock()
 
         files = {
             self.console_sock: '',
--- a/unit/test_tags.py
+++ b/unit/test_tags.py
@@ -1,0 +1,123 @@
+#!/usr/bin/env python3
+
+import unittest
+from pathlib import Path
+
+from ansible.inventory.manager import InventoryManager
+from ansible.parsing.dataloader import DataLoader
+from ansible.playbook.play import Play
+from ansible.vars.manager import VariableManager
+
+
+class TestTags(unittest.TestCase):
+    def setUp(self):
+        loader = DataLoader()
+        inventory = InventoryManager(loader=loader, sources='localhost,')
+        variable_manager = VariableManager(loader=loader, inventory=inventory)
+
+        self.play = Play().load({
+            'name': "Deploy Cartridge",
+            'tasks': [{
+                'name': 'Import Cartridge Role',
+                'import_role': {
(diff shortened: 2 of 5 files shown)


dokshina:
Please, mention issue that is closed by this PR. Also you need to update CHANGELOG.

BTW, it is possible to test tags?

unit/instance.py
```diff
@@ -38,12 +38,25 @@ def __init__(self, console_sock, cluster_cookie):
         self.process = None
         self.sock = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
 
+        self.__original_exists = os.path.exists
```
dokshina, line 41: It's not clear now why do you save original functions and then restore them. Looks that this functions should be mocked only in tests that uses them, or here in `__init__` function. I think second approach is better, because generally we use role tasks on remote host and there is no cases when we check existence of files on local machine.

hackallcode: I restore the original functions because they change not only for the current test, but also for all next tests. This may lead to the fact that in some tests the original function should be used, however, it will use the mocked function.

Also, this functions mocked in `start` function, because it was originally there. But if you want, I'll transfer it to the `__init__` function

unit/test_tags.py
```diff
@@ -0,0 +1,123 @@
...
+            'Manage failover',
+        ])
+
+    def test_not_cartridge_tag(self):
+        names = self.get_tasks_by_tags({'test'})
+        self.assertEqual(names, [
+            # TODO: Remove 'Restart instances and reload systemd-daemon' when Ansible will be updated to 2.11 or higher
+            # Ref: https://docs.ansible.com/ansible/latest/collections/ansible/builtin/meta_module.html#notes
+            'Restart instances and reload systemd-daemon',
+        ])
```
dokshina, line 123: It's very-very good that now tags are covered by tests, thanks!

But I worry about copy-paste of tasks names - each change of task name will lead to many changes. I suggest grouping tasks names in variables. Smth like 

```python
COMMON_TASKS = [
            'Check distribution',
            'Set remote_user for delegated tasks',
            'Validate config',
            'Set instance facts',
]

INSTANCE_TASKS = [
            'Install package',
            'Start instance',
            'Restart instances and reload systemd-daemon',
            'Wait for instance to be started',
]

REPLICASETS_TASKS = ...
```

and in tests

```python
self.assertEqual(names, COMMON_TASKS + INSTANCE_TASKS)
```

I think that it would be more clear than tasks lists.

hackallcode: However, some tasks have multiple tags, but not all. For example, `Select one not expelled instance` has the tags `cartridge-replicasets` and `cartridge-config` and does not have the tag `cartridge-instances`. If you want to group them, we should use a dictionary like this:
```python
{
   "Select one not expelled instance": ["cartridge-replicasets", "cartridge-config"],
}
```
And then make lists for each tag from this dictionary.
But this variant is not as readable as the current one. We will also duplicate the Ansible logic, which is not a good idea.

hackallcode: So this is a good test for checking the logic of tags without calculation, which is more correct for tests.

dokshina (approved):

PR #171, opened 2021-02-08
Add manual gathering facts task

### LINKED ISSUE #170: [2pt] Impossible to use one of tags when use in playbook
I have several tasks in my playbook. One of them is importing the anisble-cartridge role.
I made a tag for this task like this:
```yaml
- name: Deploy Tarantool Cartridge app
  hosts: cartridge_group
  tags: cartridge
  tasks:
    - name: Import Tarantool Cartridge role
       import_role:
         name: tarantool.cartridge
```
After that, I start the playbook with the `-t cartridge-instances` parameter. I expected the role will be execute tasks with `cartridge-instances` tag, but in the end I get error like this:
```
fatal: [host_name]: FAILED! => {"msg": "The conditional check 'ansible_os_family not in [\"RedHat\", \"Debian\"]' failed. The error was: error while evaluating conditional (ansible_os_family not in [\"RedHat\", \"Debian\"]): 'ansible_os_family' is undefined\n\nThe error appears to be in '.ansible/roles/tarantool.cartridge/tasks/main.yml': line 2, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n---\n- name: Check distribution\n  ^ here\n"}
```

This is because I want to use `tags: cartridge` for task with role import. Therefore, gathering of facts occurs only if the `cartridge` tag is specified.

Let's make a facts gathering task that will be completed if the facts were not gathered earlier and one of our tags is specified.

hackallcode:
Closes #170

Before the patch, if you run the playbook with the -t 'cartridge-instances' option,
you get an error like this: `'ansible_os_family' is undefined`.

Added facts gathering task will be executed if the facts were not
gathered earlier and one of our tags is specified.

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -1,4 +1,12 @@
 ---
+- name: Forced facts gathering
+  setup:
+  when: ansible_os_family is not defined
+  tags:
+    - cartridge-instances
+    - cartridge-replicasets
+    - cartridge-config
+
 - name: Check distribution
   fail:
     msg: 'Deploy to {{ ansible_os_family }} distributions is not supported yet'
--- a/unit/test_tags.py
+++ b/unit/test_tags.py
@@ -41,6 +41,7 @@ class TestTags(unittest.TestCase):
     def test_without_tags(self):
         names = self.get_tasks_by_tags({})
         self.assertEqual(names, [
+            'Forced facts gathering',
             'Check distribution',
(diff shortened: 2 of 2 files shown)


dokshina (approved):
Please, shortly describe a problem in the merge commit.

PR #172, opened 2021-02-08
Speed up "Install package" task

### LINKED ISSUE #155: [2pt] Too long "Install package" task
I have 100 Tarantool instances in a cluster. When I deploy them via ansible-cartridge, it takes a very long time to complete the "Install package" task, since it is performed for each instance. Moreover, this task is performed even despite the fact that it is written in the logs that task is skipped for all instances.

Let's define the host list once instead of a hundred times.

hackallcode:
Closes #155

### Changes

Before this patch, instances for the "Install Package" task were selected on each instance, it was too long.

Now the list of instances for installing a package is selected once for all.
This made it possible to speed up the task at times for a large cluster.

### Complexity and Time

Before this patch, the complexity of calculating the list of instances was `O(N^2)`, now it is `O(N)`.
For 100 instances, this gives a 10x time reduction, from 60 seconds to 5 seconds.


tasks/main.yml
```diff
@@ -42,12 +42,18 @@
     - cartridge-replicasets
     - cartridge-config
 
+- name: Calculate hosts list
```
dokshina: Please, don't name machines `hosts`. Ansible `hosts` are instances in this role, I think we should use "machine" word to underline that its physical servers, not instances.

filter_plugins/filter_hosts.py
```diff
@@ -1,5 +1,8 @@
-def get_machine_identifier(v):
-    return v['ansible_host'] if 'ansible_host' in v else v['inventory_hostname']
+def get_machine_hostname(v):
+    if 'ansible_host' in v:
+        return v['ansible_host']
+
+    raise Exception(f'Instance {v["inventory_hostname"]} has not "ansible_host" option!')
```
dokshina, line 5: `has not` -> `hasn't`? `doesn't have`? `has not`?

I'm not sure it's wrong.

hackallcode: I think everything is right

tasks/main.yml
hackallcode: Please, check new variant of variables names

unit/test_tags.py
```diff
@@ -45,6 +45,7 @@ def test_without_tags(self):
             'Set remote_user for delegated tasks',
             'Validate config',
             'Set instance facts',
+            'Calculate machine hosts list',
```
dokshina: Get physical machines list?
Get one instance from each physical machine to install package?

dokshina (commented):
Do you have some measurements that shows advantages of a new approach?

unit/test_tags.py
hackallcode: Get one instance from each physical machine

tasks/main.yml
```diff
@@ -42,12 +42,18 @@
     - cartridge-replicasets
     - cartridge-config
 
+- name: Get one instance from each physical machine
+  set_fact:
+    machine_hosts_list: '{{ hostvars | get_one_not_expelled_instance_for_machine(play_hosts) }}'
```
dokshina: Maybe `machine_hostnames`?

dokshina:
I think this fix can be mentioned in Changelog.

tasks/main.yml
hackallcode: But it's not a hostnames (e.g. `tarantool.io`), it's names of Ansible hosts (e.g. `my_core_instance`)

hackallcode:
> Do you have some measurements that shows advantages of a new approach?

Before this patch, the complexity of calculating the list of instances was `O(N^2)`, now it is `O(N)`. For 100 instances, this gives a 10x time reduction

tasks/main.yml
dokshina: So, it's smth like `instances_to_install_package`?

hackallcode: `single_instances_for_each_machine`

dokshina (approved):

PR #173, opened 2021-02-08
Cleanup expelled instance files and services

dokshina:
After #165 stopping expelled instances systemd services and removing this instances files was lost. 
This patch adds this actions after `Edit topology` task.
It also removed redundant `install_tgz` file.


unit/test_tags.py
```diff
@@ -11,6 +11,8 @@
 
 class TestTags(unittest.TestCase):
     def setUp(self):
+        self.maxDiff = None
```
hackallcode, line 14: What are you using this code for?

dokshina: Because if lists comparison fails, pytest doesn't show full diff, but shows warning: `Diff is 664 characters long. Set self.maxDiff to None to see it.`. I used this flag to see why did tests fail.

hackallcode (approved):

PR #174, opened 2021-02-09
Refactor package installing

dokshina:
Getting package info (name and Tarantool dependency) is moved to library
module. `install_{rpm,deb}.yml` tasks are removed, all common tasks are
moved to` install_package.yml`.


library/cartridge_get_package_info.py
```diff
@@ -0,0 +1,139 @@
...
+
+
+def get_rpm_info(package_path):
+    # package name
+    cmd = ['rpm', '-qip', package_path]
+    rc, output = run_command_and_get_output(cmd)
+    if rc != 0:
+        raise Exception("failed to get RPM package info: %s" % output)
+
+    m = re.search(r'Name\s*:\s*(.*)\s+', output)
```
hackallcode: `.*` can catch any spaces, line ends, etc. Are you sure that this regex is correct? It may be better to use `[^\s]+` or `[^\n]+`. A similar question to other places with `.*` regular expression.

tasks/install_package.yml
```diff
@@ -1,16 +1,49 @@
...
+- name: Get package meta info
+  cartridge_get_package_info:
+    package_path: '{{ copied_package.dest }}'
+    app_name: '{{ cartridge_app_name }}'
+  register: package_info
+  any_errors_fatal: true
+
+- name: Debug package_info
+  debug:
+    var: package_info
```
hackallcode: Debug?

tasks/install_package.yml
```diff
@@ -1,16 +1,49 @@
...
-  tags: cartridge-instances
+  yum:
+    name: '{{ copied_package.dest }}'
+    state: present
+    update_cache: true
+  register: install_rpm
+  failed_when: install_rpm.rc != 0 and "does not update installed package" not in install_rpm.results[0]
+  notify:
+    - reload-systemd-daemon
+  when: cartridge_package_path is regex('\.rpm$')
```
hackallcode: What about to make `package_info.meta.package_type` option?

library/cartridge_get_package_info.py
dokshina: Replaced via `\S+`.

hackallcode: Are you sure the name cannot contain spaces?

hackallcode (approved):

library/cartridge_get_package_info.py
hackallcode: Are you sure the name cannot contain spaces?
If in some cases spaces can be in the name, then let's use `[^\n]+`.
If, in all cases, there are no spaces in the application name, this is fine.

PR #175, opened 2021-02-09
Check cluster issues in molecule tests

dokshina:
Check that cluster hasn't issues after bootstrapping
Change stateboard listen to `0.0.0.0`

### DIFF
--- a/molecule/default/hosts.yml
+++ b/molecule/default/hosts.yml
@@ -79,7 +79,7 @@ all:
 
         my-stateboard:
           config:
-            listen: vm1:3310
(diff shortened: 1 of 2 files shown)


molecule/default/hosts.yml
```diff
@@ -79,7 +79,7 @@ all:
 
         my-stateboard:
           config:
-            listen: vm1:3310
+            listen: 0.0.0.0:3310
```
hackallcode, line 82: Is it mandatory to install the stateboard on the machine running the tests?

hackallcode (approved):

PR #176, opened 2021-02-11
Change cartridge_check_instance_started function

### LINKED ISSUE #145: [2pt] Wait until instance becomes 'healthy' when bootstrapping big cluster
If an new replicaset does not become 'healthy' during mass join, ansible won't wait for him, and will stop pipeline

hackallcode:
Closes #145

Thanks for a draft PR #138

Before this patch, it only checked that the instances are alive.
Moreover, the stateboard instance was not checked.

Changes:
- All instances are being checked now, including the stateboard.
- The instance is expected to be not only alive, but also in the
  Unconfigured or RolesConfigured status (you can set any list of
  expected statuses by `expected_states` option).
- For routers, added an optional check that all buckets are discovered
  if cluster was bootstrapped. Your can disable this check by
  `check_buckets_are_discovered` option.

### DIFF
--- a/library/cartridge_check_instance_started.py
+++ b/library/cartridge_check_instance_started.py
@@ -4,32 +4,108 @@ from ansible.module_utils.basic import AnsibleModule
 from ansible.module_utils.helpers import ModuleRes, CartridgeException
 from ansible.module_utils.helpers import get_control_console
 
-
 argument_spec = {
     'control_sock': {'required': True, 'type': 'str'},
     'stateboard': {'required': True, 'type': 'bool'},
+    'expected_states': {'required': False, 'type': 'list', 'default': ['Unconfigured', 'RolesConfigured']},
+    'check_buckets_are_discovered': {'required': False, 'type': 'bool', 'default': True},
 }
 
 
-def check_instance_started(params):
+def check_stateboard_started(control_console):
+    box_status, err = control_console.eval_res_err('''
+        if type(box.cfg) == 'function' or box.cfg.listen == nil then
+            return nil, "box hasn't been configured"
+        end
+        return true
+    ''')
+    if not box_status:
+        return ModuleRes(success=False, msg="Stateboard is not running: %s" % err)
+
+    return ModuleRes(success=True)
+
+
+def check_instance_started(control_console, expected_states, check_buckets_are_discovered):
+    instance_state, err = control_console.eval_res_err('''
+        return require('cartridge.confapplier').get_state()
+    ''')
+    if not instance_state:
+        return ModuleRes(success=False, msg="Impossible to get state: %s" % err)
+    if instance_state not in expected_states:
+        return ModuleRes(
+            success=False,
+            msg="Instance is not in one of states: %s, it's in '%s' state" % (
+                expected_states,
+                instance_state,
+            ),
+        )
+
+    if check_buckets_are_discovered:
+        buckets_ok, err = control_console.eval_res_err('''
+            local vshard_utils = require('cartridge.vshard-utils')
+            local vshard_router = require('cartridge.roles.vshard-router')
+
+            local function check_group_buckets(group_name, group_opts)
+                if not group_opts.bootstrapped then
+                    return true
+                end
+
+                local router = vshard_router.get(group_name)
+                if router == nil then
+                    return true
+                end
+
+                local unknown_buckets = router:info().bucket.unknown
+                if unknown_buckets == 0 then
+                    return true
+                end
+
+                return nil, string.format(
+                    "%s out of %s buckets are not discovered in group '%s'",
+                    unknown_buckets,
+                    group_opts.bucket_count,
+                    group_name
+                )
+            end
+
+            local groups = vshard_utils.get_known_groups()
+            for group_name, group_opts in pairs(groups) do
+                local _, err = check_group_buckets(group_name, group_opts)
+                if err ~= nil then
+                    return nil, err
+                end
+            end
+
+            return true
+        ''')
+        if not buckets_ok:
+            return ModuleRes(success=False, msg=err)
+
+    return ModuleRes(success=True)
+
+
+def check_started(params):
     try:
         control_console = get_control_console(params['control_sock'])
-        instance_is_alive = True
-        if not params['stateboard']:
-            instance_is_alive, _ = control_console.eval_res_err('''
-                return require('membership').myself().status == 'alive'
-            ''')
+
+        if params['stateboard']:
+            return check_stateboard_started(control_console)
+        else:
+            return check_instance_started(
+                control_console,
+                params['expected_states'],
+                params['check_buckets_are_discovered'],
+            )
+
     except CartridgeException as e:
         return ModuleRes(success=False, msg=str(e))
-
-    return ModuleRes(success=instance_is_alive)
 
 
 def main():
     module = AnsibleModule(argument_spec=argument_spec)
 
     try:
-        res = check_instance_started(module.params)
+        res = check_started(module.params)
     except CartridgeException as e:
         module.fail_json(msg=str(e))
 
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -182,7 +182,7 @@
 
 - name: Bootstrap vshard
   cartridge_bootstrap_vshard:
-    control_sock: '{{ control_sock}}'
+    control_sock: '{{ control_sock }}'
   delegate_to: '{{ control_host }}'
   run_once: true
   register: bootstrap_vshard
@@ -190,6 +190,17 @@
   retries: 3
   delay: 5
   when: cartridge_bootstrap_vshard is defined and cartridge_bootstrap_vshard
+  tags: cartridge-config
+
+- name: Wait for instance to be started and buckets are discovered
+  cartridge_check_instance_started:
+    control_sock: '{{ instance_control_sock }}'
+    stateboard: '{{ stateboard }}'
+  register: check_instance
+  until: not check_instance.failed
+  retries: '{{ instance_start_timeout // 5 }}'
+  delay: 5
+  when: not expelled and cartridge_bootstrap_vshard is defined and cartridge_bootstrap_vshard
   tags: cartridge-config
 
 - name: Manage failover
--- a/unit/test_check_instance_started.py
+++ b/unit/test_check_instance_started.py
@@ -1,27 +1,40 @@
 # Hack ansible.module_utils.helpers import
 import sys
+
 import module_utils.helpers as helpers
+
 sys.modules['ansible.module_utils.helpers'] = helpers
 
 import os
+
 sys.path.append(os.path.dirname(__file__))
 
 import unittest
 from instance import Instance
 
-from library.cartridge_check_instance_started import check_instance_started
+from library.cartridge_check_instance_started import check_started
 
 
-def call_check_instance_started(control_sock, stateboard=False):
-    return check_instance_started({
+def call_check_instance_started(control_sock, stateboard=False, check_buckets_are_discovered=False):
+    return check_started({
         'control_sock': control_sock,
         'stateboard': stateboard,
+        'expected_states': ['Unconfigured', 'RolesConfigured'],
+        'check_buckets_are_discovered': check_buckets_are_discovered,
     })
 
 
-def set_myself(instance, status):
-    instance.set_variable('membership_myself', {
-        'status': status
+def set_confapplier_state(instance, state):
+    instance.set_variable('cartridge_confapplier_state', state)
+
+
+def set_vshard_groups(instance, vshard_groups):
+    instance.set_variable('vshard_groups', vshard_groups)
+
+
+def set_vshard_router_unknown_buckets(instance, groups_unknown_buckets):
+    instance.set_variable('unknown_buckets', {
+        g[0]: g[1] for g in groups_unknown_buckets.items()
     })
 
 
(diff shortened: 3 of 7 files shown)


library/cartridge_check_instance_started.py
```diff
@@ -4,25 +4,77 @@
...
+                    return nil, string.format(
+                        '%s out of %s buckets are not discovered!',
+                        router.info().bucket.unknown,
+                        router.bucket_count()
+                    )
+                ''')
+                if not buckets_discovered:
+                    return ModuleRes(success=False, msg=err)
+
     except CartridgeException as e:
```
dokshina, line 100: CartridgeException is caught in `main` function, why do you catch it here?

hackallcode: I didn't change this code, but I'll remove it

library/cartridge_check_instance_started.py
```diff
@@ -4,32 +4,93 @@
 from ansible.module_utils.helpers import ModuleRes, CartridgeException
 from ansible.module_utils.helpers import get_control_console
 
-
 argument_spec = {
     'control_sock': {'required': True, 'type': 'str'},
     'stateboard': {'required': True, 'type': 'bool'},
+    'expected_state': {'required': False, 'type': 'list', 'default': ['Unconfigured', 'RolesConfigured']},
+    'check_buckets_discovery': {'required': False, 'type': 'bool', 'default': False},
```
dokshina: `check_buckets_discovering`?
`check_buckets_are_discoveed`?

library/cartridge_check_instance_started.py
```diff
@@ -4,32 +4,93 @@
...
-                return require('membership').myself().status == 'alive'
+def check_stateboard_started(control_console):
+    box_status, err = control_console.eval_res_err('''
+        if type(box.cfg) == 'function' then
+            return nil, "Box hasn't been configured!"
+        end
+        return box.info.status
+    ''')
+    if not box_status:
+        return ModuleRes(success=False, msg="Impossible to get stateboard status: %s!" % err)
```
dokshina: Maybe

```python
if err is not None:
    return ModuleRes(success=False, msg=err)
```

library/cartridge_check_instance_started.py
```diff
@@ -4,32 +4,93 @@
...
+    if box_status != 'running':
+        return ModuleRes(success=False, msg="Stateboard is not running, it has '%s' status!" % box_status)
+
+    return ModuleRes(success=True)
+
+
+def check_instance_started(control_console, expected_state, check_buckets_discovery):
+    member_status, err = control_console.eval_res_err("return require('membership').myself().status")
```
dokshina: Are you sure that it's possible to get `nil, err` here?

library/cartridge_check_instance_started.py
```diff
@@ -4,32 +4,93 @@
...
+    if not member_status:
+        return ModuleRes(success=False, msg="Impossible to get membership status: %s!" % err)
+    if member_status != 'alive':
+        return ModuleRes(success=False, msg="Membership is not alive, it has '%s' status!" % member_status)
```
dokshina: It's not `membership status`, it's `instance status`.

library/cartridge_check_instance_started.py
```diff
@@ -4,32 +4,93 @@
...
+
+    cartridge_state, err = control_console.eval_res_err("return require('membership').myself().payload.state")
```
dokshina: Seems that you can get `require('membership').myself()`, save this value and then check status and state without one more call.

dokshina: `cartridge_state` -> `instance_state`

library/cartridge_check_instance_started.py
```diff
@@ -4,32 +4,93 @@
...
+    if not cartridge_state:
+        return ModuleRes(success=False, msg="Impossible to get instance Cartridge state: %s!" % err)
+    if cartridge_state not in expected_state:
```
dokshina: `expected_state` or `expected_stateS`?

library/cartridge_check_instance_started.py
```diff
@@ -4,32 +4,93 @@
...
+        return ModuleRes(
+            success=False,
+            msg="Instance is not in one of states: %s, it's in '%s' state!" % (
```
dokshina: Why your errors are so emotional?) You actively use `!`.

library/cartridge_check_instance_started.py
hackallcode: If any exception is thrown

library/cartridge_check_instance_started.py
```diff
@@ -4,32 +4,93 @@
...
+
+    if check_buckets_discovery:
+        buckets_discovered, err = control_console.eval_res_err('''
+                local router = require('vshard.router')
+                if router.info == nil then
+                    return true
+                end
+
+                if require('cartridge.vshard-utils').can_bootstrap() then
+                    return nil, "Cluster isn't bootstrapped yet!"
```
dokshina: Seems that this case the same as above `shouldn't be discovered`.

Since we return `true` for `shouldn't be discovered` or `discovered successfully`, `buckets_discovered` can be renamed to `buckets_ok`, what do you think?

library/cartridge_check_instance_started.py
hackallcode: Why not, fixed

library/cartridge_check_instance_started.py
hackallcode: Yes, it's more readble

library/cartridge_check_instance_started.py
hackallcode: If any exception occurs, then we will not lose the error message and immediately understand what is wrong.

library/cartridge_check_instance_started.py
hackallcode: Okay, fixed

library/cartridge_check_instance_started.py
hackallcode: Yes, I made one call for both (status and status)

library/cartridge_check_instance_started.py
hackallcode: Yes, it's better

library/cartridge_check_instance_started.py
hackallcode: Because I think that with `.` are informational messages, and with `!` - error messages. But I removed the `!` from all messages so they don't look threatening

library/cartridge_check_instance_started.py
hackallcode: Yes, I made `buckets_ok`

library/cartridge_check_instance_started.py
```diff
@@ -4,32 +4,94 @@
...
-        if not params['stateboard']:
-            instance_is_alive, _ = control_console.eval_res_err('''
-                return require('membership').myself().status == 'alive'
+def check_stateboard_started(control_console):
+    box_status, err = control_console.eval_res_err('''
+        if type(box.cfg) == 'function' then
+            return nil, "Box hasn't been configured"
+        end
+        return box.info.status
+    ''')
```
rosik: I'd suggest to check `box.cfg.listen ~= nil`. See the `stateboard.cfg()` implementation: https://github.com/tarantool/cartridge/blob/master/cartridge/stateboard.lua#L386-L387. It also overlaps `box.info.status` check

library/cartridge_check_instance_started.py
```diff
@@ -4,32 +4,94 @@
...
+    if box_status != 'running':
+        return ModuleRes(success=False, msg="Stateboard is not running, it has '%s' status" % box_status)
+
+    return ModuleRes(success=True)
+
+
+def check_instance_started(control_console, expected_states, check_buckets_are_discovered):
+    instance_info, err = control_console.eval_res_err('''
+        local myself = require('membership').myself()
+        return { myself.status, myself.payload and myself.payload.state }
```
rosik: Checking `require('cartridge.confapplier').get_state()` is more reliable. And I think OperationError would also be acceptable for the ansible.

library/cartridge_check_instance_started.py
```diff
@@ -4,32 +4,94 @@
...
+                    return true
+                end
+
+                if require('cartridge.vshard-utils').can_bootstrap() then
+                    return nil, "Cluster isn't bootstrapped yet"
+                end
+
+                if router.info().bucket.unknown == 0 then
+                    return true
+                end
```
rosik: This logic isn't compatible with a multi-group setup.

rosik (commented):
Please, see the 3 comments below. I didn't review the tests yet, will get back to them later, when discussions in code are resolved.

library/cartridge_check_instance_started.py
```diff
@@ -4,32 +4,97 @@
...
+
+    return ModuleRes(success=True)
+
+
+def check_instance_started(control_console, expected_states, check_buckets_are_discovered):
+    instance_state, err = control_console.eval_res_err('''
+        return require('cartridge.confapplier').get_state()
+    ''')
+    if err is not None:
+        return ModuleRes(success=False, msg="Impossible to get state: %s" % err)
```
rosik: `get_state()` can `return "OperationError, err" too. It wouldn't correspond to the message "Impossible to get state".

library/cartridge_check_instance_started.py
```diff
@@ -4,32 +4,97 @@
 from ansible.module_utils.helpers import ModuleRes, CartridgeException
 from ansible.module_utils.helpers import get_control_console
 
-
 argument_spec = {
     'control_sock': {'required': True, 'type': 'str'},
     'stateboard': {'required': True, 'type': 'bool'},
+    'expected_states': {'required': False, 'type': 'list', 'default': ['Unconfigured', 'RolesConfigured']},
```
rosik, line 10: And yet, what about the OperationError?

library/cartridge_check_instance_started.py
```diff
@@ -4,32 +4,97 @@
...
+                    return nil, "Cluster isn't bootstrapped yet"
+                end
+
+                local vshard_router = require('cartridge.roles.vshard-router')
+
+                local groups = vshard_utils.get_known_groups()
+                for group_name, group_opts in pairs(groups) do
+                    local router = vshard_router.get(group_name)
+                    if router == nil then
+                        return true
```
rosik: Too quick. I guess it should be `goto continue`

library/cartridge_check_instance_started.py
hackallcode: OperationError isn't a success state. Is not it so?

library/cartridge_check_instance_started.py
```diff
@@ -4,32 +4,94 @@
...
+            success=False,
+            msg="Instance is not in one of states: %s, it's in '%s' state" % (
+                expected_states,
+                instance_state,
+            ),
+        )
+
+    if check_buckets_are_discovered:
+        buckets_ok, err = control_console.eval_res_err('''
+                local vshard_utils = require('cartridge.vshard-utils')
```
dokshina: Seems that there is 8 spaces indent.

hackallcode: Yes, I'll fix it

rosik (approved):

rosik (approved):

library/cartridge_check_instance_started.py
```diff
@@ -4,32 +4,91 @@
...
+    if check_buckets_are_discovered:
+        buckets_ok, err = control_console.eval_res_err('''
+            local vshard_utils = require('cartridge.vshard-utils')
+            local vshard_router = require('cartridge.roles.vshard-router')
 
-    return ModuleRes(success=instance_is_alive)
+            local groups = vshard_utils.get_known_groups()
+            for group_name, group_opts in pairs(groups) do
+                if group_opts.bootstrapped == true then
+                    local router = vshard_router.get(group_name)
```
dokshina: Seems that this part can be moved to some `check_group_buckets` function. Just to decrease number of nesting levels.

tasks/main.yml
```diff
@@ -192,6 +192,17 @@
   when: cartridge_bootstrap_vshard is defined and cartridge_bootstrap_vshard
   tags: cartridge-config
 
+- name: Wait for instance to be bootstrapped
```
dokshina: Let's say "Wait for instance to be STARTED". "Bootstrapped" confuses a lot - we bootstrap vshard, cluster is bootstrapped when replicasets are created. Also we use `instance_START_timeout` here.

unit/test_check_instance_started.py
```diff
@@ -1,27 +1,43 @@
...
+
 
+def set_confapplier_state(instance, state):
+    instance.set_variable('cartridge_confapplier_state', state)
 
-def set_myself(instance, status):
-    instance.set_variable('membership_myself', {
-        'status': status
+
+def set_vshard_groups_bucket_count(instance, vshard_groups):
```
dokshina: `set_vshard_groups_bucket_count` -> `set_vshard_groups`?

unit/test_check_instance_started.py
```diff
@@ -1,27 +1,43 @@
...
+    instance.set_variable('vshard_groups', vshard_groups)
+
+
+def set_vshard_router_unknown_buckets(instance, vshard_groups):
+    instance.set_variable('unknown_buckets', {
+        g[0]: g[1] for g in vshard_groups.items()
```
dokshina: Why does `set_vshard_router_unknown_buckets` accept `vshard_groups` argument?

library/cartridge_check_instance_started.py
```diff
@@ -4,32 +4,98 @@
...
+
+    if check_buckets_are_discovered:
+        buckets_ok, err = control_console.eval_res_err('''
+            local vshard_utils = require('cartridge.vshard-utils')
+            local vshard_router = require('cartridge.roles.vshard-router')
 
-    return ModuleRes(success=instance_is_alive)
+            local function check_group_buckets(group_name, bucket_count)
+                local router = vshard_router.get(group_name)
+                if router ~= nil then
```
dokshina: It can be simplified more, just use early return here.

`if router == nil then return true end`.

dokshina (approved):
Great!

PR #177, opened 2021-02-15
Refactor instances facts

dokshina:
* Each module returns values without `meta` sub-table
* Now all instance facts (like package type, console socket path configuration file path and so on) 
are placed in `instance_info` dictionary. It's computed at the play beginning by `cartridge_get_instance_info` module.
* Getting one non-expelled instance, one instance for each machine and control instance are moved to library
modules. For one non-expelled instance and control instance `console_sock` is returned as well as `console_sock`.
This values can be used in the following tasks without additional filters:

```yaml
- cartridge_get_one_not_expelled_instance:
    hostvars: '{{ hostvars }}'
    play_hosts: '{{ play_hosts }}'
  register: not_expelled_instance

- some_task:
    console_sock: '{{ not_expelled_instance.console_sock }}'
  delegate_to: '{{ not_expelled_instance.name }}'
```
 
I sure that this approach should simplify using role sub-tasks after splitting the monolith.


module_utils/helpers.py
```diff
@@ -7,7 +7,7 @@
 
 
 class ModuleRes:
-    def __init__(self, success, msg=None, changed=False, meta=None, warnings=[]):
+    def __init__(self, success, msg=None, changed=False, meta=dict(), warnings=[]):
```
hackallcode: It's a bad practice to use mutable values as defaults. Use `None` value instead.
https://docs.python-guide.org/writing/gotchas/

library/cartridge_get_control_instance.py
```diff
@@ -0,0 +1,78 @@
...
+    ''')
+
+    for _, member in members.items():
+        if 'payload' in member and member['payload'].get('uuid') is not None:
+            if member['payload'].get('alias') is None:
+                errmsg = 'Unable to get instance alias for "{}"'.format(member['payload']['uuid'])
+                return ModuleRes(success=False, msg=errmsg)
+
+            instance_name = member['payload']['alias']
+            control_instance_name = instance_name
```
hackallcode:
```suggestion
            control_instance_name = member['payload']['alias']
```

library/cartridge_get_instance_from_each_machine.py
```diff
@@ -0,0 +1,61 @@
...
+        machine_hostname = get_machine_hostname(instance_vars, instance_name)
+        if machine_hostname not in machine_hostnames:
+            machine_hostnames.add(machine_hostname)
+            instance_names.append(instance_name)
+
+    instances = {
+        'names': instance_names,
+    }
+
+    return ModuleRes(success=True, meta=instances)
```
hackallcode:
```suggestion
    return ModuleRes(success=True, meta={
        'names': instance_names,
    })
```

module_utils/helpers.py
dokshina: Please, see my fix.

module_utils/helpers.py
```diff
@@ -7,11 +7,11 @@
...
 
 class ModuleRes:
-    def __init__(self, success, msg=None, changed=False, meta=dict(), warnings=[]):
+    def __init__(self, success, msg=None, changed=False, meta=None, warnings=None):
         self.success = success
         self.msg = msg
         self.changed = changed
-        self.meta = meta
+        self.meta = meta if meta is not None else dict()
         self.warnings = warnings
```
hackallcode, line 15: Final default value is changed from `[]` to `None`, is it ok? Maybe you want to do like this:
```suggestion
        self.warnings = warnings if warnings is not None else []
```

unit/test_get_one_instance_for_machine.py
```diff
@@ -0,0 +1,67 @@
...
+        ])
+        self.assertTrue(res.success, res.msg)
+        self.assertEqual(res.meta['names'], ['stateboard-1', 'stateboard-2'])
+
+    def test_instances_found(self):
+        res = call_get_one_not_expelled_instance_for_machine(self.hostvars, [
+            'expelled-1', 'expelled-2', 'instance-1', 'instance-2', 'stateboard-1', 'stateboard-2',
+        ])
+        self.assertTrue(res.success, res.msg)
+        self.assertEqual(res.meta['names'], ['instance-1', 'instance-2'])
```
hackallcode: Maybe it's better to check that instance OR stateboard will be taken from each machine?

unit/test_get_one_non_expelled_instance.py
```diff
@@ -0,0 +1,53 @@
...
+
+    def test_instance_found(self):
+        res = call_get_one_not_expelled_instance(self.hostvars, [
+            'expelled-1', 'my-stateboard', 'expelled-2', 'instance-1', 'instance-2', 'instance-3',
+        ])
+        self.assertTrue(res.success, res.msg)
+        self.assertEqual(res.meta, {
+            'name': 'instance-1',
+            'console_sock': 'sock-1',
+        })
```
hackallcode: Maybe it's better to use `assertIn(res.meta, [instance-1, instance-2, instance-3])`?

hackallcode (approved):

PR #181, opened 2021-02-17
Refactor testinfra tests

dokshina:
Before this patch `test_services_status_and_config` test hasn't test anything because instances from machine wasn't collected.
Now machine instances are collected correctly.

### DIFF
--- a/molecule/default/tests/test_cartridge.py
+++ b/molecule/default/tests/test_cartridge.py
@@ -1,6 +1,5 @@
 import os
-import yaml
-# import copy
+from yaml import CLoader as Loader
 
 import testinfra.utils.ansible_runner
 import requests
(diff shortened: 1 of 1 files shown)


molecule/default/tests/test_cartridge.py
```diff
@@ -133,55 +130,65 @@ def aliases_in_priority_order(replicaset_servers):
...
+        conf_dir = instance_vars.get('cartridge_conf_dir', '/etc/tarantool/conf.d')
+        run_dir = instance_vars.get('cartridge_run_dir', '/var/run/tarantool')
+        data_dir = instance_vars.get('cartridge_data_dir', '/var/lib/tarantool')
+
+        if not instance_is_stateboard(instance_vars):
+            service_name = '%s@%s' % (APP_NAME, instance_name)
+            instance_id = '%s.%s' % (APP_NAME, instance_name)
+        else:
+            service_name = '%s-stateboard' % APP_NAME
+            instance_id = '%s-stateboard' % APP_NAME
```
hackallcode:
```suggestion
            instance_id = service_name = '%s-stateboard' % APP_NAME
```

molecule/default/tests/test_cartridge.py
```diff
@@ -17,7 +16,13 @@
 APP_NAME = 'myapp'
 HOSTS_PATH = os.path.join('molecule', 'default', 'hosts.yml')
 
+inventory = InventoryManager(loader=DataLoader(), sources=HOSTS_PATH)
+variable_manager = VariableManager(loader=DataLoader(), inventory=inventory)
+
+cluster_cookie = inventory.groups['cluster'].get_vars()['cartridge_cluster_cookie']
+
 __authorized_session = None
+__configured_instances = None
```
hackallcode, line 25: Why do some variables start with "__" and some don't? All of them are not used in other modules, is it right?

dokshina: This is variables used on lazy-initialization.

hackallcode (approved):

PR #183, opened 2021-02-19
Fix Tarantool installation in GitHub Actions

hackallcode:
The `installer.sh` script no longer installs Tarantool.
It's necessary to install Tarantool separately.

It was decided to use a ready-made solution for installing Tarantool.


dokshina (approved):
Brilliant!

PR #196, opened 2021-02-25
Update .gitignore file

hackallcode:
Some files were not ignored, which could lead to uploading unnecessary files to the server.

Now all unnecessary files are ignored.


dokshina (approved):
Bless!

PR #200, opened 2021-02-26
Validate TGZ and multiversion variables

dokshina:
Multiversion and TGZ options are added to variables types schema.


library/cartridge_validate_config.py
```diff
@@ -142,6 +143,16 @@ def validate_types(vars):
         'weight': int,
         'vshard_group': str,
         'cartridge_enable_tarantool_repo': bool,
+        'cartridge_install_dir': str,
+        'cartridge_instances_dir': str,
+        'cartridge_conf_dir': str,
+        'cartridge_run_dir': str,
+        'cartridge_data_dir': str,
```
hackallcode, line 150: Is it possible to check that it's not a simple string, but a path?

dokshina: I think that string is enough.

hackallcode: It seems to me that it is worth thinking about how to make a check for all paths in the config (not in this pull request).

hackallcode (dismissed):

hackallcode (approved):
Great!

PR #202, opened 2021-03-01
Remove temporary files

### LINKED ISSUE #187: [1pt] Remove temporary files
Some temporary files are delivered to hosts and are not removed from it. For example, `tarantool-installer.sh` or package.

Let's remove temporary files! Maybe it's better to add some flag, e.g. `cartridge_remove_temp_files`

hackallcode:
Closes #187

Now it's possible to remove temporary files (such as a delivered package) after use.
To do this, you should set the `cartridge_remove_temporary_files` flag to true.
Also, it's possible to add any files to `temporary_files` list by config or custom step.

### DIFF
--- a/README.md
+++ b/README.md
@@ -162,6 +162,8 @@ Common variables:
 * `cartridge_app_name` (`string`, required): application name;
 * `cartridge_cluster_cookie` (`string`, required): cluster cookie for all
   cluster instances;
+* `cartridge_remove_temporary_files` (`boolean`, optional, default: `false`): indicates if temporary
+  files should be removed (more details in description of `cleanup` [step API](/doc/scenario.md#cleanup));
 
 Role scenario configuration:
 
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -1,5 +1,8 @@
 ---
 
+temporary_files: []
+cartridge_remove_temporary_files: false
+
 cartridge_package_path: null
 cartridge_defaults: {}
 cartridge_enable_tarantool_repo: true
@@ -44,6 +47,7 @@ cartridge_scenario:
   - configure_app_config
   - bootstrap_vshard
   - configure_failover
+  - cleanup
 
 cartridge_configure_systemd_unit_files: true
 cartridge_systemd_dir: /etc/systemd/system
--- a/doc/scenario.md
+++ b/doc/scenario.md
@@ -445,8 +445,9 @@ Input facts (set by config):
 
 ### cleanup
 
-Remove temporary files from `temporary_files` list. By default, `temporary_files` is an empty list. The role, depending
-on the scenario, can add the following files to this list:
+Removes temporary files specified in `temporary_files` list.
+By default, `temporary_files` is an empty list. The role,
+depending on the scenario, can add the following files to this list:
 
 - path to delivered package (`delivered_package_path` variable value);
 - path to repository setup script.
--- a/doc/scenario.md
+++ b/doc/scenario.md
@@ -25,6 +25,7 @@ In `cartridge_scenario` you should specify names of steps. The default scenario 
 - [configure_app_config](#configure_app_config)
 - [bootstrap_vshard](#bootstrap_vshard)
 - [configure_failover](#configure_failover)
+- [cleanup](#cleanup)
 
 There are additional steps that are not included in the default scenario, but can be used in a custom one:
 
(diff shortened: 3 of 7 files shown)


README.md
```diff
@@ -175,6 +175,8 @@ Configuration format is described in detail in the
 * `cartridge_enable_tarantool_repo` (`boolean`, optional, default: `true`):
   indicates if the Tarantool repository should be enabled (for packages with
   open-source Tarantool dependency);
+* `cartridge_remove_temp_files` (`boolean`, optional, default: `false`):
```
dokshina: What is temp files for user? Maybe `cartridge_cleanup_package_file`?

BTW, there is one more tmp file we need to remove - repository setup script.

hackallcode: The script has already been removed.

Temporary files are temporary files created by a role, but these files are not needed by the application.

tasks/steps/blocks/update_package.yml
```diff
@@ -25,6 +25,10 @@
       command: bash /tmp/tarantool-installer.sh
       changed_when: false
 
+    - name: Add repository setup script to temporary files list
```
dokshina, line 28: Let's reduce a copy-paste:

```yaml
    - name: 'Get repository setup script'
      get_url:
        url: 'https://tarantool.io/release/{{ package_info.tnt_version }}/installer.sh'
        dest: '/tmp'
      register: enable_repo_script
      ...
    - name: 'Run repository setup script'
      ...
      command: bash {{ enable_repo_script.dest }}
      ...
    - name: Add repository setup script to temporary files list
      set_fact:
        temporary_files: "{{ temporary_files + [enable_repo_script.dest] }}"
```

README.md
```diff
@@ -175,6 +175,8 @@ Configuration format is described in detail in the
 * `cartridge_enable_tarantool_repo` (`boolean`, optional, default: `true`):
   indicates if the Tarantool repository should be enabled (for packages with
   open-source Tarantool dependency);
+* `cartridge_remove_temporary_files` (`boolean`, optional, default: `false`):
+  indicates if temporary files should be removed;
```
dokshina: Maybe we should notice that user can add his own temporary files to `tmpfiles_list`?

README.md
```diff
@@ -175,6 +175,8 @@ Configuration format is described in detail in the
 * `cartridge_enable_tarantool_repo` (`boolean`, optional, default: `true`):
   indicates if the Tarantool repository should be enabled (for packages with
   open-source Tarantool dependency);
+* `cartridge_remove_temporary_files` (`boolean`, optional, default: `false`):
+  indicates if temporary files should be removed (you can add your own files to `temporary_files` list);
```
dokshina: What files are in `temporary_files` by default?

defaults/main.yml
```diff
@@ -44,11 +54,4 @@ cartridge_scenario:
...
   - bootstrap_vshard
   - configure_failover
-
-cartridge_configure_systemd_unit_files: true
-cartridge_systemd_dir: /etc/systemd/system
-
-cartridge_configure_tmpfiles: true
-cartridge_tmpfiles_dir: /usr/lib/tmpfiles.d/
-
-cartridge_install_tarantool_for_tgz: true
```
dokshina: All these changes look like a separated PR...

doc/scenario.md
```diff
@@ -315,3 +316,13 @@ Input facts (set by role):
 Input facts (set by config):
 - [DEPRECATED] `cartridge_failover` - indicates if eventual failover should be enabled or disabled;
 - `cartridge_failover_params` - failover parameters.
+
+### cleanup
+
+Clean up temporary files.
```
dokshina: Which files are temporary?

doc/scenario.md
```diff
@@ -439,3 +440,28 @@ Input facts (set by config):
 
 - [DEPRECATED] `cartridge_failover` - indicates if eventual failover should be enabled or disabled;
 - `cartridge_failover_params` - failover parameters.
+
+### cleanup
+
+Remove temporary files from `temporary_files` list. By default, `temporary_files` is an empty list. The role, depending
```
dokshina: It doesn't remove files from list.

`Removes temporary files specified in `temporary_files` list.`

tasks/steps/cleanup.yml
```diff
@@ -0,0 +1,10 @@
+---
+
+- name: 'Remove temporary files'
+  file:
+    path: '{{ filepath }}'
+    state: absent
+  loop_control:
+    loop_var: filepath
+  with_items: '{{ temporary_files }}'
+  when: cartridge_remove_temporary_files
```
dokshina, line 10: Maybe it would be better to run this task for one instance per machine?

Would it work for smth like "{{ temporary_files + ['<some-instance-specific-name>'] }}"?

dokshina:
Also please update Changelog.

tasks/steps/cleanup.yml
hackallcode: > Maybe it would be better to run this task for one instance per machine?

Yes.

> Would it work for smth like "{{ temporary_files + [''] }}"?

For what? Seems that empty string is an incorrect path.

dokshina: I've mistaken in comment, sorry.

I mean that we can run this task for one instance per machine, BUT in this case `"{{ temporary_files + ['<file special for instance>'] }}"` wouldn't work, or no?

For example, user can say `"{{ temporary_files + [ '{{ inventory_hostname }}.tmp '] }}"`.

hackallcode: It's possible use different lists for each instance, so it's impossible to run this task for each machine instead instance

dokshina (approved):

PR #203, opened 2021-03-01
Rotate dists

### LINKED ISSUE #190: [2pt] Add step to rotate TGZ

dokshina:
Closes #190

### DIFF
--- a/doc/multiversion.md
+++ b/doc/multiversion.md
@@ -46,6 +46,12 @@ and statebord.
     myapp-stateboard -> {{ cartridge_app_install_dir }}/myapp-1.0.0-0
   ```
 
+## Rotaing distributions
+
+Each new version is added to `cartridge_app_install_dir` and sometimes old version
+become redundant.
+To simply rotate distributions use [`rotate_dists` step](/doc/scenario.md#rotate_dists)
+
 ## Example
 
 Let's imagine that we already have `myapp-1.0.0-0` installed:
@@ -106,4 +112,37 @@ but `instance-2` continue to use `myapp-1.0.0-0`:
   myapp.storage-1 -> {{ cartridge_app_install_dir }}/myapp-2.0.0-0
   myapp.storage-2 -> {{ cartridge_app_install_dir }}/myapp-2.0.0-0
   myapp.core-1 -> {{ cartridge_app_install_dir }}/myapp-1.0.0-0
+```
+
+Sometimes we have a lot of distributions in `cartridge_app_install_dir`:
+
+```bash
+{{ cartridge_app_install_dir }}/
+  myapp-1.0.0-0/
+  myapp-2.0.0-0/
+  myapp-3.0.0-0/
+  myapp-4.0.0-0/
+  myapp-4.0.0-0/
+```
+
+Let's rotate them:
+
+```yaml
+- name: Update storages
+  hosts: all
+  tasks:
+    - import_role:
+        name: tarantool.cartridge
+      vars:
+        cartridge_dists_retention_num: 2
+        cartridge_scenario:
+          - rotate_dists
+```
+
+After running this playbook, there are only 2 last distributions:
+
+```bash
+{{ cartridge_app_install_dir }}/
+  myapp-4.0.0-0/
+  myapp-4.0.0-0/
 ```
--- a/doc/multiversion.md
+++ b/doc/multiversion.md
@@ -122,7 +122,7 @@ Sometimes we have a lot of distributions in `cartridge_app_install_dir`:
   myapp-2.0.0-0/
   myapp-3.0.0-0/
   myapp-4.0.0-0/
-  myapp-4.0.0-0/
+  myapp-5.0.0-0/
 ```
 
 Let's rotate them:
@@ -144,5 +144,5 @@ After running this playbook, there are only 2 last distributions:
 ```bash
 {{ cartridge_app_install_dir }}/
   myapp-4.0.0-0/
-  myapp-4.0.0-0/
+  myapp-5.0.0-0/
 ```
--- a/doc/multiversion.md
+++ b/doc/multiversion.md
@@ -134,7 +134,7 @@ Let's rotate them:
     - import_role:
         name: tarantool.cartridge
       vars:
-        cartridge_dists_retention_num: 2
+        cartridge_keep_num_latest_dists: 2
         cartridge_scenario:
           - rotate_dists
 ```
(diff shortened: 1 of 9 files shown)


doc/multiversion.md
```diff
@@ -107,3 +113,36 @@ but `instance-2` continue to use `myapp-1.0.0-0`:
...
+
+Sometimes we have a lot of distributions in `cartridge_app_install_dir`:
+
+```bash
+{{ cartridge_app_install_dir }}/
+  myapp-1.0.0-0/
+  myapp-2.0.0-0/
+  myapp-3.0.0-0/
+  myapp-4.0.0-0/
+  myapp-4.0.0-0/
```
hackallcode:
```suggestion
  myapp-5.0.0-0/
```

doc/multiversion.md
```diff
@@ -107,3 +113,36 @@ but `instance-2` continue to use `myapp-1.0.0-0`:
...
+        cartridge_scenario:
+          - rotate_dists
+```
+
+After running this playbook, there are only 2 last distributions:
+
+```bash
+{{ cartridge_app_install_dir }}/
+  myapp-4.0.0-0/
+  myapp-4.0.0-0/
```
hackallcode:
```suggestion
  myapp-5.0.0-0/
```

library/cartridge_set_dist_dirs_to_remove.py
```diff
@@ -0,0 +1,58 @@
...
+argument_spec = {
+    'app_name': {'required': False, 'type': 'str'},
+    'app_install_dir': {'required': True, 'type': 'str'},
+    'dists_retention_num': {'required': True, 'type': 'int'},
+}
+
+if pkgutil.find_loader('ansible.module_utils.helpers'):
+    import ansible.module_utils.helpers as helpers
+else:
+    import module_utils.helpers as helpers
```
hackallcode: Let's use imports before variables definition

library/cartridge_validate_config.py
```diff
@@ -505,6 +506,18 @@ def validate_config(params):
...
+        if 'cartridge_dists_retention_num' in host_vars:
+            dists_retention_num = host_vars['cartridge_dists_retention_num']
+            if dists_retention_num <= 0:
+                errmsg = '"cartridge_dists_retention_num" should be greater than 0'
+                return helpers.ModuleRes(failed=True, msg=errmsg)
+            if dists_retention_num < 2:
+                warnings.append(
+                    'Using "cartridge_dists_retention_num" < 2 can be dangerous. '
+                    'Make sure that there are no instances that use old versions'
+                )
```
hackallcode: It's true only in one case: num==1. Let's make condition more readable!

library/cartridge_set_dist_dirs_to_remove.py
```diff
@@ -0,0 +1,58 @@
...
+    return True
+
+
+def get_dist_dirs_to_remove(params):
+    app_name = params['app_name']
+    dists_retention_num = params['dists_retention_num']
+    app_install_dir = params['app_install_dir']
+
+    if dists_retention_num <= 0:
+        return helpers.ModuleRes(failed=True, msg='"dists_retention_num" should be > 0')
```
hackallcode: But we have the same check in validate config step, for what it?

library/cartridge_set_dist_dirs_to_remove.py
```diff
@@ -0,0 +1,58 @@
...
+
+    dists = list(filter(
+        lambda filename: is_dist(filename, app_install_dir, app_name),
+        os.listdir(app_install_dir)
+    ))
+
+    dists = sorted(
+        dists,
+        key=lambda filename: os.path.getmtime(os.path.join(app_install_dir, filename)),
```
hackallcode, line 44: Is it possible that 1.0.0 will be newer than 2.0.0?

doc/scenario.md
```diff
@@ -29,6 +29,7 @@ In `cartridge_scenario` you should specify names of steps. The default scenario
 There are additional steps that are not included in the default scenario, but can be used in a custom one:
 
 - [set_control_instance](#set_control_instance)
+- [rotate_dists](#rotate_dists)
```
hackallcode, line 32: What about to include this step in default scenario and make flag to cleanup old packages?

doc/scenario.md
```diff
@@ -439,3 +441,18 @@ Input facts (set by config):
...
+When [multiversion approcah](/doc/multiversion.md) is used, each new application
+version is added to `cartridge_app_install_dir`.
+This step removes redundant distribution.
+
+Input facts (set by config):
+
+- `cartridge_app_name` - application name;
+- `cartridge_app_install_dir` - path to directory where application distributions
+  are placed;
+- `cartridge_dists_retention_num` - number of dists that should be kept.
```
hackallcode: Output facts:
- `dists_dirs_to_remove` - ...

tasks/steps/rotate_dists.yml
```diff
@@ -0,0 +1,7 @@
+---
+
+- name: 'BLOCK : Rotate distributions'
+  include_tasks: 'blocks/rotate_dists.yml'
+  when:
+    - inventory_hostname in single_instances_for_each_machine
```
hackallcode:
```suggestion
  when: inventory_hostname in single_instances_for_each_machine
```

tasks/steps/rotate_dists.yml
```diff
@@ -0,0 +1,7 @@
...
+  tags: cartridge-instances
```
hackallcode, line 6: Is it correct tag?

library/cartridge_set_dist_dirs_to_remove.py
dokshina: I think that this additional check can prevent removing all dists if validation doesn't work.

library/cartridge_set_dist_dirs_to_remove.py
dokshina: I don't know, but mtime allows us to be sure always.

tasks/steps/rotate_dists.yml
dokshina: Yes, why not?

doc/scenario.md
dokshina: I'm not sure, it is used only for TGZ + multiversion like `update_instance`, but `update_instance` is required to start application and for other flow it's just skipped. `rotate_dists` is some optional action that can be performed separately from deploy, and I think that it's convenient way to use it in separated playbook only when you want to remove redundant packages.

tasks/steps/rotate_dists.yml
hackallcode: If it's a separated step, maybe it's better to remove any tags, but i'm not sure

hackallcode: If this is a separate step, it might be better to remove any tags, but I'm not sure

hackallcode (approved):

PR #207, opened 2021-03-02
Rename install dir option

dokshina:
- Rename `cartridge_install_dir` -> `cartridge_app_install_dir`
- Rename `cartridge_instances_dir` -> `cartridge_app_instances_dir`


defaults/main.yml
```diff
@@ -19,7 +19,7 @@ cartridge_multiversion: false
 instance_start_timeout: 60
 instance_discover_buckets_timeout: 60
 
-cartridge_install_dir: /usr/share/tarantool
+cartridge_app_install_dir: /usr/share/tarantool
 cartridge_instances_dir: /usr/share/tarantool
```
hackallcode: What about to rename this variable? cartridge_instances_app_dir?

dokshina: I like `cartridge_instances_dir` name. It's simple and clear.

hackallcode (approved):

PR #211, opened 2021-03-03
Fix and improve `needs_restart`

### LINKED ISSUE #189: [2pt] Fix and improve needs_restart for role with scenario
In the current realization `cartridge_needs_restart` is placed in `configure_instance` step. 
It checks that instance code is newer that socket (package was updated after instance start) AND that instance configuration changes require instance restart to be applied.

This approach is wrong: in case of `[update_instance, restart_instance]` scenario we skip this check and instance wouldn't be restarted.

I suggest such approach:

* User can force or disable instance restart by using `restarted` flag. By default this flag is `null`
* On `restart_instance` step we check `restarted is not false or needs_restart is defined and needs_restart`
* `needs_restart` variable is recalculated on each step that can require instance restart:
  * `update_package` step: `needs_restart = true if package was updated`
  * `configure_instances` step: `needs_restart = needs_restart or (<configuration needs restart>) `
  * ...user can enrich this variable by adding more steps

as a result `needs_restart` works like a handler - it is `true` if any step needs restart to apply changes.

hackallcode:
Closes #189

In the current realization `cartridge_needs_restart` is placed in `configure_instance` step.
It checks that instance code is newer that socket (package was updated after instance start)
AND
that instance configuration changes require instance restart to be applied.

This approach is wrong: in case of [update_instance, restart_instance] scenario
we skip this check and instance wouldn't be restarted.

Now `needs_restart` variable is recalculated on each step that can require instance restart:
- `update_package`
- `update_instance`
- `configure_instance`

### DIFF
--- a/doc/scenario.md
+++ b/doc/scenario.md
@@ -199,12 +199,15 @@ Install the delivered package on physical machines.
 
 Input facts (set by role):
 
+- `instance_info` - information for a current instance ([more details here](#role-facts-descriptions));
 - `single_instances_for_each_machine` - list of instances (Ansible hosts), one for each physical machine;
-- `delivered_package_path` - remote path to file of delivered package.
+- `delivered_package_path` - remote path to file of delivered package;
+- `needs_restart` - if instance should be restarted to apply code or configuration changes.
 
 Input facts (set by config):
 
-- `instance_info` - information for a current instance ([more details here](#role-facts-descriptions));
+- `expelled` - indicates if instance must be expelled from topology;
+- `restarted` - if instance should be restarted or not (user forced decision);
 - `cartridge_app_name` - application name;
 - `cartridge_enable_tarantool_repo` - indicates if Tarantool repository should be enabled;
 - `cartridge_install_tarantool_for_tgz` - indicates if Tarantool should be enabled when use TGZ package;
@@ -239,7 +242,8 @@ Output facts:
   - `instance_entrypoint` - path to Lua entrypoint file of instance;
   - `stateboard_entrypoint` - path to Lua entrypoint file of stateboard;
   - `instance_tarantool_binary` - path Tarantool binary file of instance;
-  - `stateboard_tarantool_binary` - path Tarantool binary file of stateboard.
+  - `stateboard_tarantool_binary` - path Tarantool binary file of stateboard;
+- `needs_restart` - if instance should be restarted to apply code or configuration changes.
 
 ### update_instance
 
@@ -248,14 +252,21 @@ Update instance links for a new version of package (if
 
 Input facts (set by role):
 
-- `instance_info` - information for a current instance ([more details here](#role-facts-descriptions)).
+- `instance_info` - information for a current instance ([more details here](#role-facts-descriptions));
+- `needs_restart` - if instance should be restarted to apply code or configuration changes.
 
 Input facts (set by config):
 
+- `expelled` - indicates if instance must be expelled from topology;
+- `restarted` - if instance should be restarted or not (user forced decision);
 - `cartridge_multiversion` - indicates that
   [multiversion approach](/doc/multiversion.md) is enabled;
 - `cartridge_app_user` - user which will own the links;
 - `cartridge_app_group` - group which will own the links.
+
+Output facts:
+
+- `needs_restart` - if instance should be restarted to apply code or configuration changes.
 
 ### configure_instance
 
@@ -263,7 +274,8 @@ Configure instance in runtime and change instance config.
 
 Input facts (set by role):
 
-- `instance_info` - information for a current instance ([more details here](#role-facts-descriptions)).
+- `instance_info` - information for a current instance ([more details here](#role-facts-descriptions));
+- `needs_restart` - if instance should be restarted to apply code or configuration changes.
 
 Input facts (set by config):
 
@@ -333,13 +345,13 @@ Find some instance that is already in cluster. If there is no such instance, use
 Input facts (set by role):
 
 - `not_expelled_instance` - information about one not expelled instance ([more details here](#role-facts-descriptions)).
+- `instance_info` - information for a current instance ([more details here](#role-facts-descriptions)).
 
 Input facts (set by config):
 
 - `expelled` - indicates if instance must be expelled from topology;
 - `stateboard` - indicates that the instance is a stateboard;
 - `replicaset_alias` - replicaset alias, will be displayed in Web UI;
-- `instance_info` - information for a current instance ([more details here](#role-facts-descriptions)).
 
 Output facts:
 
--- a/doc/scenario.md
+++ b/doc/scenario.md
@@ -202,12 +202,15 @@ Input facts (set by role):
 - `instance_info` - information for a current instance ([more details here](#role-facts-descriptions));
 - `single_instances_for_each_machine` - list of instances (Ansible hosts), one for each physical machine;
 - `delivered_package_path` - remote path to file of delivered package;
-- `needs_restart` - if instance should be restarted to apply code or configuration changes.
+- `needs_restart` - if instance should be restarted to apply code or configuration changes
+  (to determine if it's should be checked if instance should be restarted).
 
 Input facts (set by config):
 
-- `expelled` - indicates if instance must be expelled from topology;
-- `restarted` - if instance should be restarted or not (user forced decision);
+- `expelled` - indicates if instance must be expelled from topology
+  (to determine if it's should be checked if instance should be restarted);
+- `restarted` - if instance should be restarted or not (user forced decision)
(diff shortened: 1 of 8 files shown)


library/cartridge_set_needs_restart.py
```diff
@@ -9,86 +9,21 @@
     import module_utils.helpers as helpers
 
 argument_spec = {
```
dokshina, line 11: What about `cartridge_set_needs_restart_package_updated.py` or smth like this?

library/cartridge_set_needs_restart_to_configure.py
```diff
@@ -0,0 +1,155 @@
...
+        # it mean that instance should be restarted to apply change
+        if new_value is not None:
+            if current_cfg.get(param_name) != new_value:
+                return helpers.ModuleRes(facts={'needs_restart': True})
+
+    return helpers.ModuleRes(changed=False)
+
+
+if __name__ == '__main__':
+    helpers.execute_module(argument_spec, needs_restart_to_configure)
```
dokshina: Could it be one module that works that has two modes - checking config changes and checking package update?

tasks/steps/update_instance.yml
```diff
@@ -1,11 +1,20 @@
...
-    src: '{{ instance_info.dist_dir }}'
-    dest: '{{ instance_info.instance_dist_dir }}'
-    owner: '{{ cartridge_app_user }}'
-    group: '{{ cartridge_app_group }}'
-    state: link
-  when: cartridge_multiversion
+- when: cartridge_multiversion
   tags: cartridge-instances
+  block:
+    - name: 'Update instance application'
```
dokshina: 'Update instance application version'?

tasks/steps/update_instance.yml
```diff
@@ -1,11 +1,20 @@
...
+      file:
+        src: '{{ instance_info.dist_dir }}'
+        dest: '{{ instance_info.instance_dist_dir }}'
+        owner: '{{ cartridge_app_user }}'
+        group: '{{ cartridge_app_group }}'
+        state: link
+
+    - name: 'Check if instance restart is required to update application'
```
dokshina: 'Check if instance restart is required to apply changes'?

tasks/steps/update_instance.yml
```diff
@@ -1,11 +1,20 @@
...
+      cartridge_set_needs_restart:
+        console_sock: '{{ instance_info.console_sock }}'
+        instance_dist_dir: '{{ instance_info.instance_dist_dir }}'
```
dokshina: It can be useful to pass `restarted` to this function: if restarted is not none we can simply return `needs_restart: null`

tasks/steps/configure_instance.yml
```diff
@@ -8,17 +8,19 @@
         console_sock: '{{ instance_info.console_sock }}'
         config: '{{ config }}'
         cartridge_defaults: '{{ cartridge_defaults }}'
-      when: not restarted
+      when: (restarted == false) or (restarted is none and not needs_restart)
 
-    - name: 'Check if instance restart is forced or required to apply changes'
-      cartridge_set_needs_restart:
+    - name: 'Check if instance restart is required to update config'
+      cartridge_set_needs_restart_to_configure:
```
dokshina: How does it work when `needs_restart` is `true` after updating package but configuration isn't changed?

tasks/steps/update_instance.yml
hackallcode: Then the user will not be able to understand `needs_restart == true`, because there is a restarted flag or because the code/config has been updated

tasks/steps/update_instance.yml
hackallcode: To apply what changes?

tasks/steps/update_instance.yml
hackallcode: But you are updating not version of application, but application

tasks/steps/configure_instance.yml
hackallcode: If package was updated, then `needs_restart is not none (== true)`, so this task won't be started. If package wasn't updated (or `update_package` step wasn't started), then `needs_restart is none`, so this task will check config

library/cartridge_set_needs_restart_to_configure.py
hackallcode: Yes, I made it

doc/scenario.md
```diff
@@ -198,12 +198,15 @@ Install the delivered package on physical machines.
 
 Input facts (set by role):
 
+- `instance_info` - information for a current instance ([more details here](#role-facts-descriptions));
 - `single_instances_for_each_machine` - list of instances (Ansible hosts), one for each physical machine;
-- `delivered_package_path` - remote path to file of delivered package.
+- `delivered_package_path` - remote path to file of delivered package;
+- `needs_restart` - if instance should be restarted to apply code or configuration changes.
```
dokshina: It confuses, add a note that this value would be set if it wasn't set before.

library/cartridge_set_needs_restart.py
```diff
@@ -156,6 +158,38 @@ def needs_restart(params):
...
+            helpers.cartridge_errcodes.INSTANCE_IS_NOT_STARTED_YET
+        ]
+        if e.code in allowed_errcodes:
+            return helpers.ModuleRes(facts={'needs_restart': True})
+
+        raise e
+
+    if params['check_package_updated']:
+        res = check_needs_restart_to_update_package(params)
+        if res is not None:
```
dokshina: What cases are `res is None`?

tasks/steps/configure_instance.yml
```diff
@@ -8,17 +8,20 @@
...
         config: '{{ config }}'
         cartridge_defaults: '{{ cartridge_defaults }}'
         cluster_cookie: '{{ cartridge_cluster_cookie }}'
         stateboard: '{{ stateboard }}'
         instance_info: '{{ instance_info }}'
-      when: restarted is none
+        check_config_updated: true
+      when:
+        - restarted is none
+        - needs_restart is none
```
dokshina: But if `needs_restart` is already `true` we shouldn't check it again.

unit/test_set_needs_restart.py
```diff
@@ -112,8 +147,12 @@ def test_code_was_updated(self):
...
         self.instance.set_path_m_time(self.console_sock, self.instance.DATE_YESTERDAY)
 
+        # no check
         res = call_needs_restart(console_sock=self.console_sock)
+        self.assertFalse(res.failed, msg=res.msg)
+        self.assertFalse(res.changed)
 
+        res = call_needs_restart(console_sock=self.console_sock, check_package_updated=True)
         self.assertFalse(res.failed, msg=res.msg)
         self.assertTrue(res.changed)
```
dokshina, line 167: Please, add a case when code wasn't updated.

library/cartridge_set_needs_restart.py
```diff
@@ -181,16 +182,20 @@ def needs_restart(params):
         raise e
 
     if params['check_package_updated']:
-        res = check_needs_restart_to_update_package(params)
-        if res is not None:
-            return res
+        res, err = check_needs_restart_to_update_package(params)
```
dokshina:
```python
needs_restart, err = check_needs_restart_to_update_package(params)
...
if needs_restart:
    return helpers.ModuleRes(facts={'needs_restart': True})
```

unit/test_set_needs_restart.py
```diff
@@ -143,6 +143,16 @@ def test_box_cfg_is_function(self):
         self.assertTrue('needs_restart' in res.facts and res.facts['needs_restart'] is True)
 
     def test_code_was_updated(self):
+        # code was updated yesterday, socket today - needs restart
```
dokshina: restart isn't needed

library/cartridge_set_needs_restart.py
```diff
@@ -154,10 +155,48 @@ def needs_restart(params):
...
+
+    if params['check_config_updated']:
+        needs_restart, err = check_needs_restart_to_update_config(params, control_console)
+        if err is not None:
+            return helpers.ModuleRes(failed=True, msg=err)
+        if needs_restart:
+            return helpers.ModuleRes(facts={'needs_restart': True})
 
-    return helpers.ModuleRes(changed=False)
+    return helpers.ModuleRes(changed=False, facts={'needs_restart': False})
```
dokshina, line 198: `changed=False` seems to be redundant here.

dokshina (approved):
Nice!

Please, check one comment and feel free to merge.

library/cartridge_set_needs_restart.py
hackallcode: No, because by default it's true

PR #223, opened 2021-03-05
Speed up `connect_to_membership` step

### LINKED ISSUE #146: [1pt] Excessive probes lead to super-slow joins on big cluster
Post-mortem after 500-nodes cluster deployment

hackallcode:
Closes #146

Before the patch, the difficulty of performing `connect_to_membership` step was `N^2`.
For 100 instances, the step took about 900 seconds.

Now the complexity has decreased to N, so for 100 instances the execution time is about 5 seconds.


library/cartridge_connect_to_membership.py
```diff
@@ -14,31 +14,31 @@
...
+def connect_to_membership(params):
     control_console = helpers.get_control_console(params['console_sock'])
     hostvars = params['hostvars']
     play_hosts = params['play_hosts']
 
-    for i, instance_vars in hostvars.items():
-        if 'config' not in instance_vars:
-            continue
-
+    for instance_name, instance_vars in hostvars.items():
```
olegrok, line 34: It's not review but it seems to be reasonable to use [asyncio](https://docs.python.org/3/library/asyncio.html)-based sockets.

hackallcode: For what?

olegrok: When you perform huge amount of calls it's faster to make them asynchronous than synchronous.
It's just like make set of netbox calls sequentially in single fiber or perform each call in separate fiber/using is_async=true.

hackallcode: asyncio is quite difficult to use and it seems that the main process should be initially started using asyncio. Also sometimes we have Python < 3.4 (e.g. 2.7.5 on standard Centos), and there are no asyncio.

So, I will use Lua fibers.

library/cartridge_connect_to_membership.py
```diff
@@ -16,8 +16,41 @@
 
 def probe_server(control_console, uri):
```
dokshina, line 17: `uri` -> `uriS`,  but better `advertise_uris`.

library/cartridge_connect_to_membership.py
```diff
@@ -16,8 +16,41 @@
...
+                local _, err = cartridge.admin_probe_server(advertise_uri)
+                channel:put({advertise_uri, err})
+            end)
+        end
+
+        local errors = {}
+        for _ = 1, count do
+            local res = channel:get(60)
+            if res == nil then
+                table.insert(errors, {nil, 'Impossible to get probe result: timeout'})
```
dokshina: `Impossible` -> `Failed`?

library/cartridge_connect_to_membership.py
```diff
@@ -16,8 +16,41 @@
...
+
+        local count = 0
+        local channel = fiber.channel()
+
+        for _, advertise_uri in ipairs(advertise_uris) do
+            count = count + 1
+
+            fiber.create(function()
+                local _, err = cartridge.admin_probe_server(advertise_uri)
+                channel:put({advertise_uri, err})
```
dokshina: Why not a table?

`{advertise_uri = advertise_uri, err = err}`

I don't like using `res[2]` below, like `local err = res[2].err or res[2]`.

library/cartridge_connect_to_membership.py
```diff
@@ -0,0 +1,100 @@
...
+            continue
+
+        if 'config' not in instance_vars or 'advertise_uri' not in instance_vars['config']:
+            continue
+
+        advertise_uri = instance_vars['config']['advertise_uri']
+        advertise_uris.append(advertise_uri)
+        instance_names[advertise_uri] = instance_name
+
+    res, errors = probe_server(control_console, advertise_uris)
```
dokshina: Let's rename this function.

What about `probe_instances` or `connect_instances`?

library/cartridge_connect_to_membership.py
```diff
@@ -0,0 +1,100 @@
...
+    if not errors:
+        return helpers.ModuleRes(changed=False)
+
+    warnings = []
+    for error in errors:
```
dokshina: The same as above - why not a table?

library/cartridge_connect_to_membership.py
```diff
@@ -0,0 +1,112 @@
...
+                channel:put({
+                    advertise_uri = advertise_uri,
+                    err = err and err.err or err,
+                })
+            end)
+        end
+
+        local errors = {}
+        for _ = 1, count do
+            local res = channel:get(60)
```
dokshina: Let's move `60` to some variable.

library/cartridge_connect_to_membership.py
```diff
@@ -0,0 +1,112 @@
...
+            if res == nil then
+                table.insert(errors, {msg = 'Failed to get probe result: timeout'})
```
dokshina: `'Failed to get probe result: timeout exceeded (%s seconds)'`?

library/cartridge_connect_to_membership.py
```diff
@@ -0,0 +1,112 @@
...
+    warnings = []
+
+    for i in range(int(math.ceil(len(advertise_uris) / BATCH_SIZE))):
+        advertise_uris_batch = advertise_uris[i * BATCH_SIZE: (i + 1) * BATCH_SIZE]
+
+        errors, warns = connect_instances(control_console, play_hosts, instance_names, advertise_uris_batch)
+        warnings += warns
+
+        if len(errors) > 0:
+            return helpers.ModuleRes(failed=True, msg=errors, warnings=warnings)
```
dokshina: Should we fail here or collect all errors?
We collect all errors for one batch, but fail immediately between batches. Is it OK?

library/cartridge_connect_to_membership.py
```diff
@@ -0,0 +1,112 @@
...
+        local advertise_uris = ...
+
+        local cartridge = require('cartridge')
+        local fiber = require('fiber')
+
+        local count = 0
+        local channel = fiber.channel()
+
+        for _, advertise_uri in ipairs(advertise_uris) do
+            count = count + 1
```
dokshina: Can we skip instances that are already in membership?

```lua
if membership.get_member(advertise_uri) == nil then
    count = count + 1 
....
end
```

library/cartridge_connect_to_membership.py
```diff
@@ -0,0 +1,112 @@
...
+            errors.append(error['msg'])
+            break
+
+        instance_name = instance_names[error['advertise_uri']]
+        msg = error['msg'] % instance_name
+
+        if instance_name in play_hosts:
+            errors.append(msg)
+        else:
+            warnings.append(msg)
```
dokshina: I'm not sure that collecting such warinings is good.
Imagine that you have 500 instances configured and start to start them with limits by 5 instances each time.
You'll have 495 warnings.

library/cartridge_connect_to_membership.py
olegrok: Also it's quite strange to use 60 for each call. In fact you have N * 60 seconds timeout.

You need something like:
```lua
local deadline = fiber.clock() + timeout

for ... do
    local get_timeout = deadline - fiber.clock()
    if get_timeout < 0 then
        get_timeout = 0
    end
    get(get_timeout)
end
```

library/cartridge_connect_to_membership.py
hackallcode: Results of trying to do multithreading:
- code gets much more complicated
- when using more than 45 fibers at a time, it is not possible to connect some instances (replies from instances do not come)
- when connected in batches of 32, it turns out 1:09 versus 1:21 on 440 instances, it's too small difference

So, let's use first variant without multithreading

library/cartridge_connect_to_membership.py
hackallcode: Yes, I added this check

dokshina (approved):
Please, update CHANGELOG (and probably describe why don't we use asyncronious approach).

PR #225, opened 2021-03-11
Rewrite cartridge mock for unit tests

dokshina:
Before this patch `cartridge` mock used for unit tests emulated
`cartridge` logic with mocked functions. But it became complicated 
to support `cartridge`-like logic. Since this patch real `cartridge`  
functions are called, but some networking logic is mocked. All called 
functions are wrapped to save argument calls (to check them in tests)
or return an error when it is necessary for tests.


.gitignore
```diff
@@ -1,6 +1,7 @@
 ### Project ###
 
 packages/
+**/.rocks
```
hackallcode: It works on any subdirectory
```suggestion
.rocks
```

.github/workflows/unit-tests.yml
```diff
@@ -32,8 +32,21 @@ jobs:
...
+      - name: Cache unit tests requirements
+        id: cache-packages
+        uses: actions/cache@v2
+        with:
+          path: 'unit/mock/.rocks'
+          key: ce-${{ matrix.tarantool-version }}-${{ hashFiles('./deps.sh') }}
+
+      - name: Install unit tests requirements
+        run: |
+          ./deps.sh
```
hackallcode:
```suggestion
        run: ./deps.sh
```

deps.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+# Call this script to install unit test dependencies
+
+set -e
+
+pushd ./unit/mock
+
+tarantoolctl rocks install cartridge 2.4.0
```
hackallcode, line 8: What about to use version from environment or argument to specify cartridge version?

library/cartridge_edit_topology.py
```diff
@@ -284,7 +284,7 @@ def get_edit_servers_params(instances_to_expel, cluster_instances):
             'expelled': True,
         }
         for instance_name in instances_to_expel
-        if instance_name in cluster_instances and cluster_instances[instance_name].get('uuid') is not None
+        if instance_name in cluster_instances and cluster_instances[instance_name].get('uuid')
```
hackallcode, line 287: Is it possible that `uuid == ''`? Or for what this patch?

unit/instance.py
```diff
@@ -8,7 +8,7 @@
 
 from unit.os_mock import OsPathExistsMock, OsPathGetMTimeMock
 
-script_abspath = os.path.realpath(
+instance_path = os.path.realpath(
```
hackallcode: `mock_path`? `instance_app_dir`?

unit/instance.py
```diff
@@ -29,10 +29,11 @@ class Instance:
     DATE_TODAY = 0
     DATE_TOMORROW = 1
 
-    def __init__(self, console_sock, cluster_cookie):
+    def __init__(self):
+        self.console_sock = os.path.join(os.getcwd(), 'x.sock')
```
hackallcode: `x.sock` is a strange name

unit/instance.py
```diff
@@ -57,13 +60,18 @@ def _connect(self):
...
         env = os.environ.copy()
         env['TARANTOOL_CONSOLE_SOCK'] = self.console_sock
 
         with open(os.devnull, 'w') as FNULL:
-            self.process = Popen(command, env=env, stdout=FNULL, stderr=FNULL)
+            self.process = Popen(command, env=env, cwd=cwd, stdout=FNULL, stderr=FNULL)
+
+        # # verbose output
+        # import sys
+        # self.process = Popen(command, env=env, cwd=cwd, stdout=sys.stdout, stderr=sys.stderr)
```
hackallcode: Debug

unit/instance.py
```diff
@@ -88,28 +96,33 @@ def recvall():
                 chunk = self.sock.recv(1024).decode()
                 # It is correct because of cmd structure: it always returns a value
                 if chunk == '':
-                    raise Exception(
-                        'Error: broken pipe. '
-                        'Probably, the instance was not bootstrapped yet to perform this operation'
-                    )
+                    errmsg = 'Error: broken pipe. ' + \
+                             'Probably, the instance was not bootstrapped yet to perform this operation'
+                    raise Exception(errmsg)
```
hackallcode: A strange change)

unit/instance.py
```diff
@@ -88,28 +96,33 @@ def recvall():
...
+end
+local args = require('json').decode('{args_encoded}')
+local ret = {{
+    load(
+        'local func, args = ... return func(unpack(args))',
+        '@eval'
+    )(func, args)
+}}
+return string.hex(require('json').encode(ret))
+'''
```
hackallcode:
```suggestion
        cmd_fmt = textwrap.dedent('''
            local function func(...)
                {func_body}
            end
            local args = require('json').decode('{args_encoded}')
            local ret = {{
                load(
                    'local func, args = ... return func(unpack(args))',
                    '@eval'
                )(func, args)
            }}
            return string.hex(require('json').encode(ret))
        ''')
```

unit/instance.py
```diff
@@ -122,17 +135,35 @@ def recvall():
...
-        except Exception:
-            raise Exception(hex_output)
+        if hex_output.startswith("error:"):
+            err = re.sub(r"error:\s+", '', hex_output)
+            raise Exception(err)
+
+        output = bytearray.fromhex(hex_output).decode('utf-8')
+
+        data = json.loads(output)
+        return data
```
hackallcode:
```suggestion
        return json.loads(output)
```

unit/instance.py
```diff
@@ -201,129 +233,75 @@ def set_path_m_time(path, m_time):
...
+    def add_replicaset(self, **kwargs):
+        self.eval_res_err('''
+            require('cartridge').internal.add_replicaset(...)
+        ''', kwargs)
+
+    def bootstrap_cluster(self):
+        res, err = self.eval_res_err('''
+            require('cartridge').internal.bootstrap_cluster()
+        ''')
+        assert err is None, err
```
hackallcode: What about to check errors in other places? Or maybe return an error?

unit/mock/init.lua
```diff
@@ -2,9 +2,17 @@
 require('strict').on()
 
 local fio = require('fio')
+local app_dir = fio.abspath(fio.dirname(arg[0]))
 
-local script_dir = fio.abspath(fio.dirname(arg[0]))
-package.path = package.path .. ';' .. script_dir .. '/?.lua'
+print('App dir set to ' .. app_dir)
```
hackallcode: Debug?

library/cartridge_edit_topology.py
dokshina: Yes, it turned out that `cartridge` returns `uuid = ''` for unjoined instances.

unit/instance.py
dokshina: Renamed to `instance.sock`.

unit/instance.py
dokshina: That was copy paste.

unit/instance.py
dokshina: I've added asserts in Lua code for internal functions. `eval_res_err` will raise exceptions for this errors with assert message.

unit/instance.py
dokshina: I've added `debug` flag to `instance.start`.

unit/instance.py
```diff
@@ -122,17 +140,34 @@ def recvall():
...
-        return ret['ret']
+        if len(data) == 0:
+            # return nil
+            data.append(None)
+
+        if len(data) == 1:
+            # return res
+            data.append(None)
+
+        # return res, err
```
hackallcode: Remove it

hackallcode (approved):
In the near future, it will be necessary to divide these tests into unit tests and integration tests.

PR #227, opened 2021-03-11
Allow to run role without gathering facts

### LINKED ISSUE #226: [1pt] Disable gethering facts by default

dokshina:
* Disable gathering facts in molecule test playbook.
* Remove task that checks OS (it used ansible_os_family fact).
* Add gather_facts: no in example playbook in README.

Closes #226

### DIFF
--- a/molecule/default/converge.yml
+++ b/molecule/default/converge.yml
@@ -5,3 +5,4 @@
     - role: ansible-cartridge
   become: true
   become_user: root
+  gather_facts: false
--- a/tasks/validate.yml
+++ b/tasks/validate.yml
@@ -11,4 +11,6 @@
 - name: 'Validate OS Family'
   fail:
     msg: 'Deploy to {{ ansible_os_family }} distributions is not supported yet'
-  when: ansible_os_family not in ["RedHat", "Debian"]
+  when:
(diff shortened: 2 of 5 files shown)


tasks/validate.yml
```diff
@@ -7,8 +7,3 @@
   run_once: true
   delegate_to: localhost
   become: false
-
-- name: 'Validate OS Family'
```
hackallcode, line 11: Let's add to README that this role available only on "RedHat", "Debian" families

molecule/default/converge.yml
```diff
@@ -5,3 +5,4 @@
     - role: ansible-cartridge
   become: true
   become_user: root
+  gather_facts: no
```
hackallcode: Let's use `false` value

no1seman: You can disable gather_facts by default,  move OS checks to separate block and call setup: explicitly

dokshina: Now OS checks are performed only if `ansible_os_family` fact is defined. 
I think that if one uses default scenario, he shouldn't change it just to check OS family.
But if facts were gathered we can additionally check OS family.

hackallcode (approved):

PR #232, opened 2021-03-12
Test role with pipelining: true

dokshina:


hackallcode:
It's pipeline before patch:
https://github.com/tarantool/ansible-cartridge/actions/runs/642238864
After patch:
https://github.com/tarantool/ansible-cartridge/actions/runs/645767635

Speed is the same

dokshina:
I've realized that we use `docker` connection in tests so `ssh_connection` option doesn't affect anything.

PR #241, opened 2021-03-18
Provide cartridge upgrade compatibility

### LINKED ISSUE #237: Cartridge upgrade compatibilty
Seems that need to investigate the case when upgrading Cartridge 2.3.0, 2.4.0 - > 2.5.0 using ansible-cartridge role. 

Refactor two-phase commit (patch_clusterwide) logics in Cartridge 2.5.0 need some special checks and possibly separate logic when upgrading from Cartridge 2.3.0, 2.4.0.

dokshina:
Since this patch on selecting control instance two-phase commit
version of all candidates (instances that are already joined or all instances
that should be joined when bootstrapping from scratch) is checked and 
instance with the lowest version is selected.

There is one rule that allows to work with cluster that has instances 
with different versions of Tarantool: **all operations that modify 
clustar-wide config should be performed via instance that has lowest
Cartridge version**. In fact, only version of two-phase commit matters, 
e.g. Cartridge 2.5.0 has incompatible change of two-phase commit and
instance with new version can't correctly update config on instances 
that uses older versions. 

We can't simply get Cartridge versions that are
used by every instance since some projects use Cartridge scm-1.
Since https://github.com/tarantool/cartridge/pull/1327 `twophase.VERSION`
is available. For older Cartridge version we can simply check if 
`_G.__cartridge_upload_begin` exists. In this case twophase version is 2, 
otherwise - 1.

Closes #237

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,9 @@ to use the newest tag with new release
   after editing topology
 - availability to specify instance `memtx_dir`, `vinyl_dir` and `wal_dir` params
   by `cartridge_memtx_dir_parent`, `cartridge_vinyl_dir_parent`, `cartridge_wal_dir_parent` variables.
+- Control instance is selected considering two-phase commit version of instances.
+  The reason is that all operations that modify cluster-wide config should be performed via instance
+  that has lowest Cartridge version (in fact, only two-phase commit version matters).
 
 ### Changed
 
--- a/library/cartridge_set_control_instance.py
+++ b/library/cartridge_set_control_instance.py
@@ -15,6 +15,69 @@ argument_spec = {
 }
 
 
+GET_TWOPHASE_COMMIT_VERSION_TIMEOUT = 60
+
+
+def get_twophase_commit_versions(control_console, advertise_uris):
+    versions, err = control_console.eval_res_err('''
+        local fiber_clock = require('fiber').clock
+        local pool = require('cartridge.pool')
+
+        local uris, timeout = ...
+
+        local deadline = fiber_clock() + timeout
+
+        local connections = {}
+        for i, uri in ipairs(uris) do
+            local conn, err = pool.connect(uri)
+            if err ~= nil then
+                return nil, tostring(err)
+            end
+
+            table.insert(connections, conn)
+        end
+
+        local futures = {}
+        for _, conn in ipairs(connections) do
+            local future = conn:eval([[
+                local twophase_version = require('cartridge.twophase').VERSION
+                if twophase_version ~= nil then
+                    return twophase_version
+                end
+
+                if rawget(_G, '__cartridge_upload_begin') ~= nil then
+                    return 2
+                else
+                    return 1
+                end
+            ]], nil, {is_async = true})
+
+            table.insert(futures, future)
+        end
+
+        local versions = {}
+
+        for _, future in ipairs(futures) do
+            local wait_timeout = deadline - fiber_clock()
+            if wait_timeout < 0 then
+                wait_timeout = 0
+            end
+
+            local result, err = future:wait_result(wait_timeout)
+            if err ~= nil then
+                return nil, tostring(err)
+            end
+
+            local version = result[1]
+            table.insert(versions, version)
+        end
+
+        return versions
+    ''', advertise_uris, GET_TWOPHASE_COMMIT_VERSION_TIMEOUT)
+
+    return versions, err
+
+
 def get_control_instance(params):
     hostvars = params['hostvars']
     play_hosts = params['play_hosts']
@@ -22,7 +85,7 @@ def get_control_instance(params):
     app_name = params['app_name']
 
     control_console = helpers.get_control_console(console_sock)
-    control_instance_name = None
+    control_instance_candidates = []
 
     members, _ = control_console.eval_res_err('''
         return require('membership').members()
@@ -44,22 +107,32 @@ def get_control_instance(params):
             if instance_name not in hostvars:
                 continue
 
-            control_instance_name = instance_name
-            break
+            control_instance_candidates.append(instance_name)
 
-    if control_instance_name is None:
+    if not control_instance_candidates:
         for instance_name in play_hosts:
             instance_vars = hostvars[instance_name]
             if helpers.is_expelled(instance_vars) or helpers.is_stateboard(instance_vars):
                 continue
 
             if 'replicaset_alias' in instance_vars:
-                control_instance_name = instance_name
-                break
+                control_instance_candidates.append(instance_name)
 
-    if control_instance_name is None:
+    if not control_instance_candidates:
         errmsg = 'Not found any joined instance or instance to create a replicaset'
         return helpers.ModuleRes(failed=True, msg=errmsg)
+
+    advertise_uris = [
+        hostvars[instance_name]['config']['advertise_uri']
+        for instance_name in control_instance_candidates
+    ]
+
+    twophase_commit_versions, err = get_twophase_commit_versions(control_console, advertise_uris)
+    if err is not None:
+        return helpers.ModuleRes(failed=True, msg=err)
+
+    idx = twophase_commit_versions.index(min(twophase_commit_versions))
+    control_instance_name = control_instance_candidates[idx]
 
     # in the ideal imagined world we could just use
     # instance_vars['instance_info'], but if control instance is not
--- a/molecule/update_cartridge/converge.yml
+++ b/molecule/update_cartridge/converge.yml
@@ -1,0 +1,91 @@
+---
+- name: Bootstrap cluster with Cartridge 2.1.2 on all instances
+  hosts: cluster
+  roles:
+    - ansible-cartridge
+  become: true
+  become_user: root
+  gather_facts: false
+  vars:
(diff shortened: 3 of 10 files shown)


library/cartridge_set_control_instance.py
```diff
@@ -15,14 +15,69 @@
...
+def get_twophase_commit_versions(control_console, advertise_uris):
+    versions, err = control_console.eval_res_err('''
+        local fiber_clock = require('fiber').clock
+        local pool = require('cartridge.pool')
+
+        local uris, timeout = ...
+        local futures = {}
+
+        for i, uri in ipairs(uris) do
+            local conn, err = pool.connect(uri)
```
hackallcode, line 32: It it possible that this connect will be too long (more than timeout)?

library/cartridge_set_control_instance.py
```diff
@@ -15,14 +15,69 @@
...
+            futures[i] = future
+        end
+
+        local versions = {}
+
+        local deadline = fiber_clock() + timeout
+        for i, future in pairs(futures) do
+            local wait_timeout = deadline - fiber_clock()
+            if wait_timeout < 0 then
+                wait_timeout = 0
```
hackallcode, line 63: return nil, 'timeout'?

library/cartridge_set_control_instance.py
```diff
@@ -15,14 +15,69 @@
...
+            }
+            local future = conn:eval([[
+                if rawget(_G, '__cartridge_upload_begin') ~= nil then
+                    return 2
+                else
+                    return 1
+                end
+            ]], nil, eval_opts)
+
+            futures[i] = future
```
hackallcode:
```suggestion
            table.insert(futures, future)
```

unit/test_set_control_instance.py
```diff
@@ -6,6 +6,20 @@
...
+
+
+def get_twophase_commit_versions_mock(_, advertise_uris):
+    versions = [
+        twophase_commit_versions.get(uri, 1) for uri in advertise_uris
+    ]
+    return versions, None
+
+
+import library.cartridge_set_control_instance as lib
```
hackallcode: Let's move this import to other imports

unit/test_set_control_instance.py
```diff
@@ -249,6 +275,73 @@ def test_instance_is_not_alive(self):
...
+                'cartridge_run_dir': RUN_DIR3,
+                'config': {'advertise_uri': URI3},
+                'replicaset_alias': 'rpl-1',
+            },
+        }
+
+        # URI2 has lower version of twophase commit
+        twophase_commit_versions[URI1] = 3
+        twophase_commit_versions[URI2] = 2
+        twophase_commit_versions[URI3] = 1
```
hackallcode:
```suggestion
        twophase_commit_versions = {
            URI1: 3,
            URI2: 2,
            URI3: 1,
        }
```

library/cartridge_set_control_instance.py
olegrok: In this case you could get two different errors for timeout.

`nil, timeout` - at this line  and `nil, box.error.new(box.error.TIMEOUT)` few lines below.

It's only reasonable here to check that you correctly handle box.errors on Python side.

unit/test_set_control_instance.py
dokshina: This import is used only for mocking function, I think that it's ok to place this logic separately.

library/cartridge_set_control_instance.py
dokshina: I've added `tostring`.

library/cartridge_set_control_instance.py
olegrok: It's possible. So "wait_connecter=false" is required here to start "connection" procedure, so then you need to iterate over all connections and call `:wait_connected()`

molecule/update_cartridge/converge.yml
```diff
@@ -0,0 +1,62 @@
+---
+- name: Bootstrap cluster with Cartridge 2.3.0 on all instances
+  hosts: cluster
+  roles:
+    - role: ansible-cartridge
```
hackallcode:
```suggestion
    - ansible-cartridge
```

molecule/update_cartridge/converge.yml
```diff
@@ -0,0 +1,62 @@
...
+  become: true
+  become_user: root
+  gather_facts: false
+  vars:
+    cartridge_scenario_name: bootstrap_cluster
+    cartridge_enable_tarantool_repo: true
+    cartridge_configure_systemd_unit_files: true
+    cartridge_configure_tmpfiles: true
+    cartridge_install_tarantool_for_tgz: true
```
hackallcode, line 13: What about to add comment that this variables is true by default?

molecule/update_cartridge/converge.yml
```diff
@@ -0,0 +1,62 @@
...
+  become: true
+  become_user: root
+  gather_facts: false
+  vars:
+    cartridge_scenario_name: update_tgz
+    cartridge_package_path: myapp-0.5.0-0-with-c-2.5.0.tar.gz
+
+# This play affects two-phase commmit
+# We need to be sure that role selects such control instance
+# tha two-phase commit doesn't fail for that
```
hackallcode:
```suggestion
# that two-phase commit doesn't fail for that
```

molecule/update_cartridge/hosts.yml
```diff
@@ -0,0 +1,85 @@
...
+          update_tgz:
+            - deliver_package
+            - update_package
+            - update_instance
+            - restart_instance
+            - wait_instance_started
+
+      # instances
+      hosts:
+        # XXX: uris lexicographic order should be reversed
```
hackallcode: XXX?)

hackallcode: Should be reversed for what?

molecule/update_cartridge/hosts.yml
```diff
@@ -0,0 +1,85 @@
+---
+all:
```
hackallcode: We don't use `all` group. What about to remove it to simplify file?

unit/test_set_control_instance.py
hackallcode: It's better to do all the imports at the top anyway:
https://www.flake8rules.com/rules/E402.html

library/cartridge_set_control_instance.py
```diff
@@ -15,14 +15,83 @@
...
+        local futures = {}
+        for _, conn in ipairs(connections) do
+            local wait_timeout = deadline - fiber_clock()
+            if wait_timeout < 0 then
+                wait_timeout = 0
+            end
+
+            if not conn:wait_connected(wait_timeout) then
+                return nil, string.format("Failed to connect to instances: timeout exceeded")
+            end
```
olegrok: No. Seem this code should be moved after the first loop.

```
        for i, uri in ipairs(uris) do
            local conn, err = pool.connect(uri, {
                wait_connected=false
            })
            if err ~= nil then
                return nil, tostring(err)
            end
            table.insert(connections, conn)
        end

          for _, conn in ipairs(connections) do
            local ok, err = conn:wait_connected(timeout)
            if not ok then
                return nil, err
            end
        end
```

molecule/update_cartridge/converge.yml
dokshina: For what? This flags are set to false in inventory to rewrite them only in first play.

molecule/update_cartridge/converge.yml
```diff
@@ -0,0 +1,62 @@
+---
+- name: Bootstrap cluster with Cartridge 2.3.0 on all instances
+  hosts: cluster
+  roles:
+    - ansible-cartridge
+  become: true
+  become_user: root
+  gather_facts: false
+  vars:
+    cartridge_scenario_name: bootstrap_cluster
```
hackallcode: Let's use default scenario

CHANGELOG.md
```diff
@@ -41,6 +41,9 @@ to use the newest tag with new release
   after editing topology
 - availability to specify instance `memtx_dir`, `vinyl_dir` and `wal_dir` params
   by `cartridge_memtx_dir_parent`, `cartridge_vinyl_dir_parent`, `cartridge_wal_dir_parent` variables.
+- Two-phase commit version of instances is considered on selecting control instance because
```
hackallcode: Select control instance use two-phase commit version...

hackallcode (commented):
Also let's remove tgz files from repository

library/cartridge_set_control_instance.py
```diff
@@ -15,14 +15,85 @@
...
+
+        local uris, timeout = ...
+
+        local deadline = fiber_clock() + timeout
+
+        local connections = {}
+        for i, uri in ipairs(uris) do
+            local conn, err = pool.connect(uri, {
+                wait_connected = false
+            })
```
rosik: I suggest calling it without options. And removing the second wait_connected loop

dokshina: Fixed.

hackallcode: But in this case, it may turn out that the connection will take longer than the timeout, isn't it?

hackallcode (approved):
It seems very good!

PR #242, opened 2021-03-18
Fix quck start example

dokshina:
Advertise URIs shown on screenshot didn't match URIs shown in config.
Now one virtual machine is used for both instances and URIs are right.


doc/files/Vagrantfile
```diff
@@ -2,12 +2,7 @@ boxes = [
     {
         :name     => "vm1",
         :ip => "172.19.0.2",
-        :ports    => [8181],
-    },
-    {
-        :name     => "vm2",
-        :ip => "172.19.0.3",
-        :ports    => [8182, 8183],
+        :ports    => [8181, 8182],
```
hackallcode, line 5: Why are we now using one machine instead of two?

hackallcode (approved):

PR #243, opened 2021-03-18
Use cartridge 2.5.0 app in tests

dokshina:
Updated Cartridge CLI to 2.7.1 in tests. It creates application with Cartridge 2.5.0.
Also create-package.sh is rewritten since applications with Cartridge 2.5.0 should
be packed in Docker (see https://github.com/tarantool/cartridge-cli/pull/503).

### DIFF
--- a/create-packages.sh
+++ b/create-packages.sh
@@ -18,13 +18,14 @@ fi
 appname=myapp
 version=1.0.0-0
 
-pack_flags=''
-if [[ $(tarantool -V) == *"Target: Darwin"* ]]; then
-    pack_flags+=' --use-docker'
+pack_flags='--use-docker'
 
-    if [[ $(tarantool -V) == "Tarantool Enterprise"* && -z "${TARANTOOL_SDK_PATH}" ]]; then
+if [[ $(tarantool -V) == "Tarantool Enterprise"* && -z "${TARANTOOL_SDK_PATH}" ]]; then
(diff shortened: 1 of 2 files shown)


create-packages.sh
```diff
@@ -18,14 +18,15 @@ fi
 appname=myapp
 version=1.0.0-0
 
-pack_flags=''
-if [[ $(tarantool -V) == *"Target: Darwin"* ]]; then
-    pack_flags+=' --use-docker'
+pack_flags='--use-docker'
```
hackallcode, line 21: Why we are using docker for linux now?

create-packages.sh
```diff
@@ -18,14 +18,15 @@ fi
...
 
-    if [[ $(tarantool -V) == "Tarantool Enterprise"* && -z "${TARANTOOL_SDK_PATH}" ]]; then
+if [[ $(tarantool -V) == "Tarantool Enterprise"* && -z "${TARANTOOL_SDK_PATH}" ]]; then
+    if [[ $(tarantool -V) == *"Target: Darwin"* ]]; then
         echo "Set the path to Linux Tarantool SDK using the TARANTOOL_SDK_PATH environment variable!"
         exit 1
     fi
+
+    pack_flags+=' --sdk-local'
```
hackallcode: For what this flag? Why was it not there before?

dokshina: Fixed!

olegrok (approved):

PR #244, opened 2021-03-19
Fix prepare scenario

dokshina:
Before this patch cartridge_scenario variable was changed
on computing scenario by cartridge_scenario_name.
This variable was passed to next plays, so if cartridge_scenario
wasn't specified, the value computed before was used instead of
computing scenario by cartridge_scenario_name.

### DIFF
--- a/doc/scenario.md
+++ b/doc/scenario.md
@@ -291,7 +291,7 @@ List of facts:
   - `name` - instance name (Ansible host);
   - `console_sock` - path to control socket of instance;
 - `cartridge_scenarios` - finally dictionary with scenarios (combination of role and user scenarios);
-- `cartridge_scenario` - steps of selected scenario (if not previously defined);
+- `scenario_steps_names` - names of scenario steps;
 - `scenario_steps` - description of scenario steps. It's a dictionary with fields:
   - `name` - name of step;
   - `path` - path to YAML file of step.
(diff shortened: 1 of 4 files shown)


tasks/prepare_scenario.yml
```diff
@@ -1,26 +1,36 @@
 ---
 
+# We should store scenario in separated variable to
+# avoid conflicts between plays.
+- name: Set scenario
```
hackallcode:
```suggestion
- name: 'Collect names of custom scenario steps'
```

tasks/prepare_scenario.yml
```diff
@@ -1,26 +1,36 @@
...
+  set_fact:
+    scenario: '{{ cartridge_scenario }}'
```
hackallcode: You should to add information about new output facts in [scenario.md](/tarantool/ansible-cartridge/blob/fix-prepare-scenario/doc/scenario.md)

doc/scenario.md
```diff
@@ -291,7 +291,7 @@ List of facts:
   - `name` - instance name (Ansible host);
   - `console_sock` - path to control socket of instance;
 - `cartridge_scenarios` - finally dictionary with scenarios (combination of role and user scenarios);
-- `cartridge_scenario` - steps of selected scenario (if not previously defined);
+- `scenario_steps_names` - names of scenario steps;
```
hackallcode, line 294: Variable `scenarios` was lost

dokshina: It was `cartridge_scenarios`. I've renamed it.

hackallcode (approved):

PR #246, opened 2021-03-22
Add debug function using warnings

hackallcode:
Now developers can debug any information using warnings!

### DIFF
--- a/module_utils/helpers.py
+++ b/module_utils/helpers.py
@@ -8,33 +8,87 @@ import textwrap
 
 from ansible.module_utils.basic import AnsibleModule
 
-
 DEFAULT_RUN_DIR = '/var/run/tarantool'
+DEBUG_MESSAGES = []
+
+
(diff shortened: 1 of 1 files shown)


module_utils/helpers.py
```diff
@@ -8,33 +8,61 @@
...
+    global DEBUG_MESSAGES
+
+    if type(value) != str:
+        value = json.dumps(value, indent=2)
+    value = value.split('\n')
+    if key:
+        value = map(lambda v: '[%s]: %s' % (key, v), value)
+
+    last_line = len(DEBUG_MESSAGES) + 1
+    value = map(lambda v: '[%03d]: %s' % (last_line + v[0], v[1]), enumerate(value))
```
dokshina, line 51: Please add a comment why do you format messages such way.

dokshina: It would be also great to show how to use `debug` function.

dokshina (approved):

PR #247, opened 2021-03-22
Remove tags unit test

dokshina:
Since tasks are imported dynamically according to scenario, we can't
test tasks without running play. Now test_tags.py doesn't make sence.


hackallcode (approved):
Indeed, they have lost their relevance!

PR #250, opened 2021-03-22
Fix error on configuring auth without users

dokshina:
Before this patch `configure_auth` step failed if user didn't specify `auth.users`.


hackallcode (approved):
Let's add description with example that doesn't work before patch

PR #258, opened 2021-03-25
Refactor GitHub actions

hackallcode:
- remove duplicate code in GitHub Actions
- transfer instances to other ports to avoid deleting the default process in Ubuntu
- update dependencies


.github/workflows/molecule-check.yml
```diff
@@ -5,7 +5,7 @@ on:
   workflow_dispatch:
 
 env:
-  CARTRIDGE_CLI_VERSION: '2.6.0'
+  CARTRIDGE_CLI_VERSION: '2.7.1'
```
dokshina: There are already CLI 2.7.2 and SDK r391.

.github/workflows/molecule-check.yml
```diff
@@ -52,6 +47,7 @@ jobs:
           git config --global user.email "test@tarantool.io" \
             && git config --global user.name "Tar Antool"
 
+          curl -L https://tarantool.io/release/${{ matrix.tarantool-version }}/installer.sh | sudo -E bash -s
```
dokshina, line 50: It's required to install cartridge-cli.

.github/workflows/molecule-tests.yml
```diff
@@ -10,145 +10,28 @@ env:
...
-
-  molecule-tests-ce-update-cartridge:
     if: |
       github.event_name == 'push' ||
       github.event_name == 'pull_request' && github.event.pull_request.head.repo.owner.login != 'tarantool'
     runs-on: ubuntu-latest
     strategy:
       matrix:
         tarantool-version: ["2.5"]
+        molecule-tag: ["default", "tasks_from", "update_cartridge"]
```
dokshina: It's `molecule-scenario`, not a tag.

.github/workflows/molecule-check.yml
dokshina: Thanks for adding =)

dokshina (approved):

PR #263, opened 2021-03-26
Use `set_fact` module instead `ansible_facts` in module result

### LINKED ISSUE #262: [2pt] Use `set_fact` module instead `ansible_facts` in module result
Now, in tasks named `cartridge_set_*`, we use the `ansible_facts` parameter to save some facts from the module. This option has a drawback: if someone previously used `set_fact` to change the value of a variable, then it's impossible to change value using `ansible_facts`.

Let's use `set_fact` to avoid variable priority conflicts.

hackallcode:
Closes #262

Before patch in tasks named `cartridge_set_*`, we used the `ansible_facts` parameter
to save some facts from the module. This option has a drawback: if someone previously
used `set_fact` to change the value of a same variable, then it's impossible to change
value using `ansible_facts`.

Now `set_fact` used to avoid variable priority conflicts.


dokshina (approved):
Please, change `Remember ...` names by `Set fact...`

PR #266, opened 2021-03-28
Use playbook variables in tasks that use `hostsvars`

### LINKED ISSUE #259: [4pt] Use playbook variables in tasks that use `hostsvars`
Some tasks use the `hostsvars` magic variable, but it does not take into account the variables from the playbook.

Let's also use the variables from the playbook in these tasks.

hackallcode:
Closes #259

Before the patch, some tasks use the `hostsvars` magic variable,
but it does not take into account the variables from the playbook.

Now the `role_vars` variable has been added in `hostsvars`,
in which all role variables are stored, taking into account the playbook
and other places where variables are declared.


library/cartridge_connect_to_membership.py
```diff
@@ -32,6 +32,9 @@ def connect_to_membership(params):
     changed = False
 
     for instance_name, instance_vars in hostvars.items():
+        if 'role_vars' in instance_vars:
```
dokshina, line 35: What if `role_vars` is not in `instance_vars`?

dokshina: Why not `instance_vars = hostvars[instance_name]['role_vars']` (as below)?

library/cartridge_get_control_instance.py
```diff
@@ -122,6 +123,8 @@ def get_control_instance(params):
         return helpers.ModuleRes(failed=True, msg=errmsg)
 
     advertise_uris = [
+        hostvars[instance_name]['role_vars']['config']['advertise_uri']
+        if 'role_vars' in hostvars[instance_name] else
```
dokshina, line 127: Why?

library/cartridge_validate_config.py
```diff
@@ -144,8 +146,6 @@ def validate_types(vars):
         'weight': int,
         'vshard_group': str,
         'cartridge_enable_tarantool_repo': bool,
-        'cartridge_install_dir': str,
-        'cartridge_instances_dir': str,
```
dokshina, line 148: Why?

tasks/prepare.yml
```diff
@@ -1,9 +1,22 @@
...
+- name: 'Get final values of role vars for each instance'
+  set_fact:
+    role_vars: "{{ vars | dict2items | selectattr('key', 'in', default_vars) | items2dict }}"
+
 - name: 'Set facts that can be set by the user'
   set_fact:
-    delivered_package_path: '{{ cartridge_delivered_package_path | default(omit) }}'
-    control_instance: '{{ cartridge_control_instance | default(omit) }}'
+    delivered_package_path: '{{ cartridge_delivered_package_path | default(omit, true) }}'
+    control_instance: '{{ cartridge_control_instance | default(omit, true) }}'
```
dokshina, line 19: I've tried to find the doc, but I didn't found description of this argument.

unit/test_connect_to_membership.py
```diff
@@ -4,9 +4,15 @@
...
-def call_probe_instance(console_sock, hostvars, play_hosts=None):
+def call_probe_instance(console_sock, role_vars, play_hosts=None):
     if play_hosts is None:
-        play_hosts = hostvars.keys()
+        play_hosts = role_vars.keys()
+
+    hostvars = {}
+    for instance_name, instance_role_vars in role_vars.items():
+        hostvars[instance_name] = {
+            'role_vars': instance_role_vars
```
dokshina, line 14: Why modules don't accept `role_vars` directly instead of `hostvars`?

library/cartridge_connect_to_membership.py
hackallcode: Because some instances are not in play hosts, so, they don't have `role_vars`

library/cartridge_validate_config.py
hackallcode: Because we don't use this variables in role

library/cartridge_get_control_instance.py
hackallcode: Up. Sometimes there are no `role_vars`

tasks/prepare.yml
hackallcode: To use the default value when variables evaluate to false or an empty string:
https://docs.ansible.com/ansible/latest/user_guide/playbooks_filters.html#providing-default-values

unit/test_connect_to_membership.py
hackallcode: Because `role_vars` - vars on one host, `hostvars` - vars on all hosts

tasks/prepare.yml
dokshina: Thanks!

dokshina (dismissed):

dokshina (approved):

PR #268, opened 2021-03-29
Add script to count test matrix

hackallcode:
Before this patch, there was GitHub Actions code duplication
as it was difficult to combine the same test with different parameters.

Now a script has been added that generates the test matrix,
so duplication of the code has been removed.

Also, most of the tests were turned off by default.
To start them, you should call the event "workflow_dispatch".


.github/helpers/count_molecule_matrix.py
```diff
@@ -0,0 +1,53 @@
+#!/usb/bin/env python3
+
+import argparse
+import json
+
+
+def get_matrix_base(ansible_version=None, molecule_command=None, molecule_scenario=None):
+    return {
+        'ansible_version': ansible_version or '2.10.0',
```
dokshina, line 9: What about declaring default values as function default args?

hackallcode: Then you will have to duplicate it in several functions.

dokshina (dismissed):
This is a fake approve - only for tests!

.github/helpers/count_molecule_matrix.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+def get_ce_version(tarantool_version=None, ansible_version=None, molecule_command=None, molecule_scenario=None):
+    return {
+        **get_matrix_base(ansible_version, molecule_command, molecule_scenario),
+        'tarantool_version': tarantool_version or '2.6',
+    }
+
+
+def get_ee_version(sdk_version=None, ansible_version=None, molecule_command=None, molecule_scenario=None):
```
dokshina: This function returns not just version, but paramters for CE job.

.github/workflows/molecule-tests.yml
```diff
@@ -55,23 +75,23 @@ jobs:
...
 
-      - name: Molecule tests
-        run: molecule test -s '${{ matrix.molecule-scenario }}'
+      - name: Molecule test
+        run: molecule ${{ matrix.molecule_command }} -s '${{ matrix.molecule_scenario }}'
 
-  molecule-tests-ee:
-    if: github.event_name == 'push'
+  molecule_tests_ee:
+    needs: setup_molecule_matrix
```
dokshina: It's not "molecule" matrix, it's "molecule test" matrix (e.g. sdk-version).

.github/workflows/molecule-tests.yml
```diff
@@ -55,23 +75,23 @@ jobs:
...
     runs-on: ubuntu-latest
+    if: needs.setup_molecule_matrix.outputs.ee_matrix_has_values == 'true'
```
dokshina: What about renaming `ee_matrix_has_values` -> `ee_tests_found`?

.github/workflows/molecule-tests.yml
```diff
@@ -4,23 +4,40 @@ name: Molecule tests
...
+            --event_name "${{ github.event_name }}" \
+            --repo_owner "${{ github.event.pull_request.head.repo.owner.login }}" \
+            --review_state "${{ github.event.review.state }}"
+    outputs:
+      ce_matrix: ${{ steps.setup_matrix.outputs.ce_matrix }}
+      ce_matrix_has_values: ${{ steps.setup_matrix.outputs.ce_matrix_has_values }}
+      ee_matrix: ${{ steps.setup_matrix.outputs.ee_matrix }}
+      ee_matrix_has_values: ${{ steps.setup_matrix.outputs.ee_matrix_has_values }}
+
+  molecule_tests_ce:
```
dokshina: Please, return kebab-case (look at GH Actions directives, like `runs-on`).

dokshina (commented):
Let's describe anywhere how do we run tests and in what order.

.github/workflows/molecule-tests.yml
hackallcode: But IDs like `pull_request_review` e.t.c. has a snake_case

dokshina (dismissed):
One more fake approve.

dokshina (approved):
Seems that this approve isn't fake.

PR #270, opened 2021-03-29
Fix test matrix generator

hackallcode:
Before this patch, on push in pull request with approve,
it was imposssible to run all tests.

Now if pull request is mergable, then all tests will be started.


dokshina (dismissed):
Fake approve.

hackallcode:
It was decided to discard approvals after pushing.

PR #277, opened 2021-04-01
Fix role vars fact

hackallcode:
Before this patch, complex variables (lists, dictionaries) were not fully expanded
on old Ansible versions. As a result, it was impossible to use other variables in
dictionaries and lists in `hosts.yml`.

Now such variables are expanded, so you can use variable value as `{{ other_var }}`.


hackallcode:
The selected approach required installing `jinja2` on servers, so it doesn't work for us

PR #281, opened 2021-04-02
Avoid using the `non_expelled_instance` fact name

hackallcode:
Before this patch some places used `non_expelled_instance` variable name, some `not_expelled_instance`.
So, it does not work correctly.

Now in all places `not_expelled_instance` is used.


dokshina (dismissed):
Please, update Changelog.

dokshina (dismissed):

dokshina (approved):
Third approve - lucky approve.

PR #301, opened 2021-04-16
Add edit topology timeouts

### LINKED ISSUE #169: Failed to edit topology: Timeout exceeded
I have a cluster of 100 instances (10 replicas with 10 instances). When deploying Cartridge role, I very often (4 times out of 5) get the next error in the "Edit Topology" step after 10-15 seconds of executing:
`fatal: [tnt_50_external_1_core_1_1 -> 10.0.0.102]: FAILED! => {"changed": false, "msg": "Failed to edit topology: Timeout exceeded"}`.

Let's make it possible to configure the maximum timeout.

hackallcode:
Closes #169

Before the patch, in a cluster of 100 instances, when deploying the cartridge role,
very often (4 times out of 5), you could receive the following error at the "Edit Topology" step
after 10-15 seconds of execution: "Failed to change topology: timed out".

Now it's possible to configure timeouts of two-phase commits on Cartridge 2.5+.


doc/scenario.md
```diff
@@ -528,7 +528,11 @@ Input facts (set by config):
 - `failover_priority` - failover priority order;
 - `all_rw` - indicates that that all servers in the replicaset should be read-write;
 - `weight` - vshard replicaset weight;
-- `vshard_group` - vshard group.
+- `vshard_group` - vshard group;
+- `edit_topology_netbox_call_timeout` - time in seconds to wait netbox call while editing topology;
```
dokshina: Note about minimum Cartridge version.

tasks/steps/edit_topology.yml
```diff
@@ -12,6 +12,9 @@
         module_hostvars: '{{ cached_facts.edit_topology }}'
         play_hosts: '{{ play_hosts }}'
         console_sock: '{{ control_instance.console_sock }}'
-        timeout: '{{ edit_topology_timeout }}'
+        netbox_call_timeout: '{{ edit_topology_netbox_call_timeout }}'
```
dokshina: Maybe we can pass it as a one table variable? Like
```
opts:
  netbox_call_timeout: '{{ netbox_call_timeout }}'
  ..
```

doc/variables.md
```diff
@@ -99,7 +99,13 @@ For more details see [scenario documentation](/doc/scenario.md).
 * `all_rw` (`boolean`): indicates that that all servers in the replicaset should be read-write;
 * `vshard_group` (`string`): name of vshard group the replicaset belongs to;
 * `weight` (`number`): vshard replicaset weight (matters only if `vshard-storage` role is enabled);
-* `edit_topology_timeout` (`number`, default: `60`): time in seconds to wait until a cluster become
+* `edit_topology_netbox_call_timeout` (`number`): time in seconds to wait netbox call
```
dokshina: Not sure that `edit_topology_` prefix is requred. Seems that `vars.options` affects everything, am I right?

module_utils/helpers.py
```diff
@@ -415,6 +429,22 @@ def get_cluster_replicasets(control_console):
...
+
+
+def set_twophase_options_from_params(control_console, params):
+    twophase_options = {}
+    if params.get('netbox_call_timeout') is not None:
+        twophase_options['netbox_call_timeout'] = params['netbox_call_timeout']
+    if params.get('upload_config_timeout') is not None:
+        twophase_options['upload_config_timeout'] = params['upload_config_timeout']
+    if params.get('apply_config_timeout') is not None:
+        twophase_options['apply_config_timeout'] = params['apply_config_timeout']
```
olegrok:
```py
allowed_options = [...]
twophase_options = {}
for option in allowed_options:
    if params.get(option) is not None:
        twophase_options[option] = params[option]
```

defaults/main.yml
```diff
@@ -104,14 +104,20 @@ roles: null
 all_rw: null
 vshard_group: null
 weight: null
-edit_topology_timeout: 60
+
+twophase_netbox_call_timeout: null
+twophase_upload_config_timeout: null
+twophase_apply_config_timeout: null
+
+edit_topology_timeout: null  # DEPRECATED
```
olegrok, line 112: I see no warnings `Option "edit_topology_timeout" is deprecated and will be removed in future release, use "edit_topology_healthy_timeout" instead`.

module_utils/helpers.py
hackallcode: The point is that specifically these options cannot be None, so we check them for None to filter.  Other options may be nullable, so they should not be checked.  Perhaps it is better to write that this is not_null_options

defaults/main.yml
hackallcode: It's a good idea to make warnings for deprecated variables!

hackallcode: It's a good idea to make warnings for deprecated variables!

olegrok (dismissed):

CHANGELOG.md
```diff
@@ -21,6 +21,15 @@ README.md to use the newest tag with new release
 - `failover_promote` step to promote replicasets leaders
 - Allowed to skip user and group creation for tgz
 - Debug control instance and one not expelled instance
+- Timeouts for edit topology step:
```
dokshina: `edit_topology` and ... steps.

doc/scenario.md
```diff
@@ -530,7 +530,15 @@ Input facts (set by config):
 - `failover_priority` - failover priority order;
 - `all_rw` - indicates that that all servers in the replicaset should be read-write;
 - `weight` - vshard replicaset weight;
-- `vshard_group` - vshard group.
+- `vshard_group` - vshard group;
+- `twophase_netbox_call_timeout` - time in seconds to wait netbox call
+  while editing topology (Cartridge 2.5+ is required);
```
dokshina: editing topology -> changing cluster-wide config?

dokshina (approved):

PR #304, opened 2021-04-20
Improve control instances debug

dokshina:
Since this patch control instance and the not-expelled instance are
mentioned on each task that uses it.

### DIFF
--- a/tasks/steps/blocks/set_control_instance.yml
+++ b/tasks/steps/blocks/set_control_instance.yml
@@ -5,7 +5,7 @@
   when: not not_expelled_instance
   run_once: true
 
-- name: 'Select control instance to manage topology and configuration'
+- name: 'Select control instance to manage topology and configuration via {{ not_expelled_instance.name }}'
   cartridge_get_control_instance:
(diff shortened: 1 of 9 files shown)


tasks/steps/blocks/set_control_instance.yml
```diff
@@ -5,7 +5,7 @@
   when: not not_expelled_instance
   run_once: true
 
-- name: 'Select control instance to manage topology and configuration'
+- name: 'Select control instance to manage topology and configuration via {{ not_expelled_instance.name }}'
```
hackallcode, line 6: Change 'Connect instance to membership' too

tasks/steps/blocks/set_control_instance.yml
```diff
@@ -5,7 +5,7 @@
   when: not not_expelled_instance
```
hackallcode, line 4: What about to return import_tasks instead include_tasks for set_not_expelled_instance?

hackallcode (approved):
Add some description for this PR and feel free to merge!

PR #310, opened 2021-04-23
Add availability to upload app config to Cartridge, TDG and more

### LINKED ISSUE #285: [4pt] Add a step to upload TDG config

hackallcode:
Closes #285 

Before this patch, the config could only be loaded by section. This led to the impossibility of loading the config into TDG or other applications that use not-lua files in their config.

Now, in addition to loading by specifying specific sections of the config, you can load the file or directory config. To do this, you should specify the path to the file or folder using the `cartridge_app_config_path` variable.

There are 3 modes for config loading (it can be specified by `cartridge_app_config_upload_mode`):
- `lua` - config will be uploaded by Lua-function `cartridge.config_patch_clusterwide()`;
- `http` - config will be uploaded to `http://localhost:port/admin/config`
  with basic authentication by cluster cookie;
- `tdg` - config will be loaded by a Lua-function in TDG with a version `1.6.15`+, `1.7.6`+ or `2.1.0`+;
  config will be loaded by HTTP in an older version.

The url used to send the config will be taken from the control instance config
(`http://127.0.0.1:{{ control_instance.http_port }}/admin/config` will be used)
or from the `cartridge_app_config_upload_url` variable. Variable
`cartridge_app_config_upload_url` takes precedence over control instance url.

Also, it's possible to specify TDG token by `cartridge_tdg_token` variable, if TDG authentication is enabled.


defaults/main.yml
```diff
@@ -115,9 +117,14 @@ edit_topology_allow_missed_instances: false
...
 
-cartridge_bootstrap_vshard: false
 cartridge_app_config: null
+cartridge_app_config_path: null
+cartridge_app_config_upload_endpoint: null
+cartridge_app_config_upload_http_port: null
+cartridge_tdg_token: null
+
 cartridge_auth: null
+cartridge_bootstrap_vshard: false
```
dokshina, line 127: Why did you place this after auth?

module_utils/helpers.py
```diff
@@ -50,6 +50,19 @@
     'vinyl_memory',
 }
 
+SYSTEM_CONFIG_SECTIONS = {
```
dokshina, line 53: Why not set?

module_utils/helpers.py
```diff
@@ -488,6 +530,51 @@ def set_twophase_options_from_params(control_console, params):
...
+    return filtered, None
+
+
+def patch_clusterwide_config(control_console, new_sections):
+    current_config, err = get_clusterwide_config(control_console)
+    if err is not None:
+        return None, err
+
+    patch = {}
+    for section_name, section in new_sections.items():
```
dokshina, line 562: What about checking all sections and show one error for all system sections?
Smth like 

```python
passed_system_sections = [
    name for name in new_sections if name in new_sections
]

if passed_system_sections:
    return None, "... %s" % ', '.join(passed_system_sections)
```

module_utils/helpers.py
```diff
@@ -488,6 +530,51 @@ def set_twophase_options_from_params(control_console, params):
...
+        if section_name in SYSTEM_CONFIG_SECTIONS:
+            return None, "Unable to patch config system section: '%s'" % section_name
+
+        if current_config.get(section_name) != section:
+            patch[section_name] = section
```
dokshina, line 568:
```python
patch = {
    section_name: section
    for section_name, section in new_sections.items()
    if current_config.get(section_name) != section:
}
```

tasks/steps/upload_app_config.yml
```diff
@@ -0,0 +1,27 @@
...
+        src: '{{ cartridge_app_config_path }}'
+        dest: '/tmp/'
+
+    - name: 'Apply application config'
+      cartridge_apply_app_config:
+        local_config_path: '{{ cartridge_app_config_path }}'
+        remote_dir: '/tmp/'
+        upload_endpoint: '{{ cartridge_app_config_upload_endpoint }}'
+        console_sock: '{{ control_instance.console_sock }}'
+        http_port: '{{ cartridge_app_config_upload_http_port or control_instance.http_port }}'
```
dokshina: Why do you allow to specify `cartridge_app_config_upload_http_port`? Because we can select control instance without HTTP enabled?

unit/test_get_control_instance.py
```diff
@@ -36,10 +36,13 @@ def call_get_control_instance(app_name, console_sock, module_hostvars=None, play
     })
 
 
-def get_instance_hostvars(alias, replicaset_alias=None, run_dir=None, expelled=False):
+def get_instance_hostvars(alias, replicaset_alias=None, run_dir=None, expelled=False, http_port=8080):
```
dokshina: Why not `http_port=None`?

library/cartridge_apply_app_config.py
```diff
@@ -0,0 +1,231 @@
...
+            raise AssertionError("Unknown TDG upload mode '%s'" % tdg_upload_mode)
+
+    else:
+        raise AssertionError("Unknown upload endpoint '%s'" % upload_endpoint)
+
+    return apply_config, dir_mode
+
+
+def apply_app_config(params):
+    remote_config_path = os.path.join(params['remote_dir'], os.path.basename(params['local_config_path']))
```
dokshina, line 219: I would like `remote_dir` to be named `tmp_dir`.

library/cartridge_apply_app_config.py
```diff
@@ -0,0 +1,231 @@
...
+    'upload_endpoint': {'required': False, 'type': 'str'},
+    'console_sock': {'required': False, 'type': 'str'},
+    'http_port': {'required': False, 'type': 'int'},
+    'cluster_cookie': {'required': False, 'type': 'str'},
+    'tdg_token': {'required': False, 'type': 'str'},
+}
+
+LUA_ENDPOINT = 'lua'
+HTTP_ENDPOINT = 'http'
+TDG_ENDPOINT = 'tdg'
```
dokshina: Why do you name it `endpoint`? It looks like it HTTP endpoint for config uploading. What about `mode`?

library/cartridge_apply_app_config.py
```diff
@@ -0,0 +1,231 @@
...
+    assert file_ext in ['.yml', '.yaml'], "Impossible to use '%s' file in patch clusterwide function" % file_ext
+
+    new_sections, err = helpers.read_yaml_file(control_console, file_path)
+    assert err is None, err
+
+    old_sections, err = helpers.get_clusterwide_config(control_console)
+    assert err is None, err
+
+    for section, value in old_sections.items():
+        new_sections[section] = new_sections.get(section)
```
dokshina, line 39:
```python
new_sections.update({
    name: new_sections.get(section)
    for section in old_sections
})
```

What if some section were specified in old config, but wasn't in new? Is it ok to simply remove it?

library/cartridge_apply_app_config.py
```diff
@@ -0,0 +1,231 @@
...
+
+    return True
+
+
+def apply_tdg_config(console_sock, path):
+    control_console = helpers.get_control_console(console_sock)
+    _, err = control_console.eval_res_err('''
+        return admin.upload_config_api(...)
+    ''', path)
+    assert err is None, err
```
dokshina, line 78: Assert?

defaults/main.yml
hackallcode: Alphabet order)

module_utils/helpers.py
hackallcode: It was list before, I'll change it

module_utils/helpers.py
hackallcode: Good idea!

module_utils/helpers.py
hackallcode: We should to check that section is not system, so it's a not good variant for it.

tasks/steps/upload_app_config.yml
hackallcode: Yes, you are right. I don't see any point in making a complex automatic system for selecting an HTTP port, since the HTTP port is almost always enabled

unit/test_get_control_instance.py
hackallcode: It doesn't affect anything, but why not

library/cartridge_apply_app_config.py
hackallcode: But this is the directory where the config is located, it is not necessarily temporary

library/cartridge_apply_app_config.py
hackallcode: It was the first version of naming, in doc I used `mode` word)

library/cartridge_apply_app_config.py
hackallcode: Yes, because it's a standard behavior of HTTP endpoint

library/cartridge_apply_app_config.py
hackallcode: Why not? It will be catched by module try-catch

tasks/steps/upload_app_config.yml
```diff
@@ -0,0 +1,27 @@
...
+- run_once: true
+  delegate_to: '{{ control_instance.name }}'
+  when: cartridge_app_config_path is not none
+  tags: cartridge-config
+  block:
+    - name: 'Copy application config on remote host'
+      any_errors_fatal: true
+      copy:
+        src: '{{ cartridge_app_config_path }}'
+        dest: '/tmp/'
```
dokshina, line 17: What about a possibility to pass directly `remote_path`? When does the user want to deliver config by himself?

CHANGELOG.md
```diff
@@ -47,6 +47,7 @@ README.md to use the newest tag with new release
   - `fencing_pause`
 - `edit_topology_allow_missed_instances` variable to allow replicasets containing
   the instances that are not started yet
+- `upload_app_config` step to load the file or directory config
```
dokshina, line 50: Maybe add a note that TDG config is also supported?

doc/app_config.md
```diff
@@ -50,3 +56,100 @@ section-1: value-1  # hasn't been changed
 section-2:
   key-21: value-21-new  # body was replaced
 ```
+
+## Config uploading
+
+In contrast to using `cartridge_app_config`, `cartridge_app_config_path`
```
dokshina: Seems strange. Maybe just the "`cartridge_upload_app_config` step (link to step doc) allows loading the entire config using HTTP endpoint or Lua API."?

doc/app_config.md
```diff
@@ -50,3 +56,100 @@ section-1: value-1  # hasn't been changed
...
+
+If you use HTTP mode to upload a file, then the file will be sent as is
+to `http://localhost:port/admin/config`. Also, in this mode will be added
+the `Authorization` header for basic authorization using a cluster cookie.
+
+If you use HTTP mode to upload a directory, it will be pre-zipped into a ZIP archive.
+This ZIP archive will be uploaded to `http://localhost:port/admin/config` just like any other file.
+
+The port used to send the config will be taken from the control instance config
+or from the `cartridge_app_config_upload_http_port` variable, if specified.
```
dokshina: Say about priority.

doc/app_config.md
```diff
@@ -50,3 +56,100 @@ section-1: value-1  # hasn't been changed
...
+
+For example, you can specify the port of the instance,
+which is on the same machine as the control instance:
```
dokshina, line 123: "on the same machine as the control" is bad. The user shouldn't know which instance is control, and he shouldn't think about it. I suggest changing `cartridge_app_config_upload_http_port` to `cartridge_app_config_upload_baseurl` to fully delegate to user setting host and port available.

doc/app_config.md
```diff
@@ -50,3 +56,100 @@ section-1: value-1  # hasn't been changed
...
+TDG mode - smart mode for loading config into TDG.
+On new versions of TDG (`1.6.15`+,` 1.7.6`+ or `2.1.0`+),
+the directory with the config will be delivered to the machine
+and then the path to the archive will be transferred to TDG by Lua.
+To do this, you do not need to specify an authorization token or HTTP port.
+
+With older versions of TDG, things are a little more complicated.
+It's possible to upload config only by HTTP.
+Therefore, if authorization is enabled in the application,
+then it is necessary to generate a token and add it to the `cartridge_tdg_token` variable:
```
dokshina: "add it" -> "pass it"

library/cartridge_apply_app_config.py
```diff
@@ -0,0 +1,244 @@
...
+        upload_mode = upload_mode.lower()
+
+    if not os.path.isdir(remote_config_path):
+        if os.path.splitext(remote_config_path)[1] in ['.yml', '.yaml']:
+            upload_mode = upload_mode or LUA_MODE
+        else:
+            upload_mode = upload_mode or HTTP_MODE
+
+        assert upload_mode in [LUA_MODE, HTTP_MODE], \
+            'File config available only for Lua or HTTP endpoints'
```
dokshina: 'Uploading file config is possible only for Lua and HTTP modes'?

library/cartridge_apply_app_config.py
```diff
@@ -0,0 +1,244 @@
...
+    else:
+        upload_mode = upload_mode or HTTP_MODE
+
+        assert upload_mode in [HTTP_MODE, TDG_MODE], \
+            'Directory config available only for HTTP or TDG endpoints'
```
dokshina: See above.

library/cartridge_apply_app_config.py
```diff
@@ -0,0 +1,244 @@
...
+
+    return upload_mode
+
+
+def get_apply_config_func_for_lua(console_sock):
+    assert console_sock is not None, "Console socket is required for Cartridge endpoint"
```
dokshina: for Lua mode?

library/cartridge_apply_app_config.py
```diff
@@ -0,0 +1,244 @@
...
+
+    control_console = helpers.get_control_console(console_sock)
+
+    return lambda path: patch_file_clusterwide(control_console, path)
+
+
+def get_apply_config_func_for_http(http_port, cluster_cookie):
+    assert cluster_cookie is not None, 'Cluster cookie is required for HTTP endpoint'
```
dokshina: HTTP mode.

dokshina: And below.

molecule/config_upload/converge.yml
```diff
@@ -0,0 +1,110 @@
+---
+
+- name: 'Test file config'
+  hosts: cluster
+  become: true
+  become_user: root
+  gather_facts: false
+  vars:
+    cartridge_eval_body: return require('cartridge').config_get_readonly()['test-section']
```
dokshina: Long line

```
cartridge_eval_body: >-
    return require('cartridge').config_get_readonly()['test-section']
```

molecule/config_upload/hosts.yml
```diff
@@ -0,0 +1,61 @@
...
+        core-1-replica:
+          config:
+            advertise_uri: 'vm1:3302'
+
+      children:
+        # group by hosts
+        machine_1:
+          vars:
+            ansible_host: vm1
+            cartridge_package_path: ./packages/myapp-1.0.0-0.rpm
```
dokshina: Move it to `all` section.

tasks/steps/upload_app_config.yml
hackallcode: I'm think it's not necessary now. We use this folder in other tasks too

CHANGELOG.md
```diff
@@ -48,6 +48,7 @@ README.md to use the newest tag with new release
 - `edit_topology_allow_missed_instances` variable to allow replicasets containing
   the instances that are not started yet
 - `upload_app_config` step to load the file or directory config
+  (Cartridge and TDG supported)
```
dokshina: are supported =)

doc/app_config.md
```diff
@@ -114,7 +115,9 @@ If you use HTTP mode to upload a directory, it will be pre-zipped into a ZIP arc
 This ZIP archive will be uploaded to `http://localhost:port/admin/config` just like any other file.
 
 The port used to send the config will be taken from the control instance config
-or from the `cartridge_app_config_upload_http_port` variable, if specified.
+or from the `cartridge_app_config_upload_http_port` variable.
+Variable `cartridge_app_config_upload_http_port` takes precedence
```
dokshina: Remove it.

library/cartridge_apply_app_config.py
```diff
@@ -151,27 +151,27 @@ def get_upload_mode(upload_mode, remote_config_path):
...
 
         assert upload_mode in [LUA_MODE, HTTP_MODE], \
-            'File config available only for Lua or HTTP endpoints'
+            'Uploading file config is possible only for Lua and HTTP modes'
     else:
         upload_mode = upload_mode or HTTP_MODE
 
         assert upload_mode in [HTTP_MODE, TDG_MODE], \
-            'Directory config available only for HTTP or TDG endpoints'
+            'Uploading directory config is possible only for HTTP or TDG endpoints'
```
dokshina: endpoints -> modes

molecule/config_upload/converge.yml
```diff
@@ -72,7 +73,7 @@
         success_msg: 'Upload folder config by Lua failed'
         that:
           - apply_app_config_res.failed
-          - apply_app_config_res.msg == 'Directory config available only for HTTP or TDG endpoints'
+          - apply_app_config_res.msg == 'Uploading directory config is possible only for HTTP or TDG endpoints'
```
dokshina: endpoints -> modes

doc/app_config.md
hackallcode: Already removed, it's outdated)

doc/app_config.md
```diff
@@ -114,15 +114,15 @@ the `Authorization` header for basic authorization using a cluster cookie.
 If you use HTTP mode to upload a directory, it will be pre-zipped into a ZIP archive.
 This ZIP archive will be uploaded to `http://localhost:port/admin/config` just like any other file.
 
-The port used to send the config will be taken from the control instance config
-or from the `cartridge_app_config_upload_http_port` variable.
-Variable `cartridge_app_config_upload_http_port` takes precedence
-over control instance HTTP port.
+The url used to send the config will be taken from the control instance config
+(`http://127.0.0.1:{control_instance.http_port}/admin/config` will be used)
```
dokshina: Maybe like Jinja? `http://127.0.0.1:{{ control_instance.http_port }}/admin/config`

dokshina (approved):
Bless you!

PR #317, opened 2021-04-27
Fix passing `force_inconsistency` option

dokshina:
Before this patch, the `force_inconsistency` flag wasn't passed correctly - 
it was passed as an argument but should be an option.

### DIFF
--- a/molecule/rolling_update/converge.yml
+++ b/molecule/rolling_update/converge.yml
@@ -117,9 +117,8 @@
   vars:
     cartridge_scenario:
       - force_leaders
-    vars:
-      cartridge_failover_promote_params:
-        force_inconsistency: false
+    cartridge_failover_promote_params:
+      force_inconsistency: false
 
 - name: 'Check new replicasets leaders'
   hosts: core-1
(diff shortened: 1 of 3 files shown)


molecule/rolling_update/converge.yml
```diff
@@ -117,6 +117,9 @@
   vars:
     cartridge_scenario:
       - force_leaders
+    vars:
+      cartridge_failover_promote_params:
+        force_inconsistency: false
```
hackallcode:
```suggestion
    cartridge_failover_promote_params:
      force_inconsistency: false
```

molecule/rolling_update/converge.yml
```diff
@@ -245,6 +248,7 @@
         core-1: core-1
         storage-1: storage-1-leader
         storage-2: storage-2-leader
+      force_inconsistency: false
```
hackallcode, line 250: What about to test force_inconsistency: true?

hackallcode (dismissed):
Add description and feel free to merge

hackallcode (approved):

hackallcode (approved):
Add description and feel free to merge

PR #318, opened 2021-04-28
Consider enabled roles on editing topology

### LINKED ISSUE #311: Do not change the topology if there are dependent roles
Now, if you specify a role in the replicaset `roles` option that is dependent on another role that is not specified in the replicaset `roles` option, then Ansible will make an unnecessary topology change.

For example, the TDG `connector` role depends on the `vshard-router` role. So, we have the `connector` role in our inventory, but the `admin_get_replicasets` function returns the `vshard-router` and `connector` roles. As result, Ansible will change topology, it's not good.

dokshina:
Before this patch passed `roles` were compared with `replicaset.roles`.
But `replicaset.roles` contain roles and their dependencies, so it can
be different from passed `roles`. This leads to unnecessary
`edit_topology` calls for all replicasets, which leads to different
issues on big clusters. Since this patch enabled roles for passed
`roles` and `replicaset.roles` are compared to understand if changes
are needed.

Closes #311

### DIFF
--- a/create-packages.sh
+++ b/create-packages.sh
@@ -41,6 +41,11 @@ awk '{gsub(/cartridge.cfg\({/, "&\n    vshard_groups = { hot = { bucket_count = 
     ${appname}/init.lua >${appname}/temp.lua
 mv ${appname}/temp.lua ${appname}/init.lua
 
+awk '{gsub(/-- dependencies/, "dependencies")}1' \
+    ${appname}/app/roles/custom.lua >${appname}/temp.lua
+mv ${appname}/temp.lua ${appname}/app/roles/custom.lua
+
+
 cartridge pack tgz --version ${version} ${pack_flags} ${appname}
 cartridge pack rpm --version ${version} ${pack_flags} ${appname}
 cartridge pack deb --version ${version} ${pack_flags} ${appname}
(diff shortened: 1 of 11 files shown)


create-packages.sh
```diff
@@ -41,6 +41,11 @@ awk '{gsub(/cartridge.cfg\({/, "&\n    vshard_groups = { hot = { bucket_count =
     ${appname}/init.lua >${appname}/temp.lua
 mv ${appname}/temp.lua ${appname}/init.lua
 
+awk '{gsub(/-- dependencies/, "dependencies")}1' \
```
hackallcode, line 44: Is it was in all versions of default app? What about to add dependencies, not uncomment?

library/cartridge_edit_topology.py
```diff
@@ -392,9 +428,11 @@ def edit_topology(params):
     control_console = helpers.get_control_console(console_sock)
 
     helpers.set_twophase_options_from_params(control_console, params)
+    set_enabled_roles(replicasets, control_console)
 
     cluster_instances = helpers.get_cluster_instances(control_console)
     cluster_replicasets = helpers.get_cluster_replicasets(control_console)
+    # roles_deps = get_roles_deps(control_console)
```
hackallcode: Remove

create-packages.sh
dokshina: Yes, it is.
I think that it's enough now - this code doesn't seem to be changed.

hackallcode (commented):
Fix conflicts and let's merge it!

hackallcode (dismissed):
LGTM!

hackallcode (approved):

PR #321, opened 2021-04-30
Remove duplicated molecule files

hackallcode:
Before the patch:
1. Molecule files were duplicated, which made it difficult to work with them.
2. Params of GutHub Actions pipelines was in uncomfortable order.

Now:
1. Most of the molecule files are in the common folder, and the tests use symbolic links to them.
2. Also, order of GutHub Actions matrix params changed.


molecule/common/molecule/1_vm_not_idempotence.yml
```diff
@@ -0,0 +1,63 @@
+---
```
dokshina: Maybe `*_no_idempotece`?

dokshina (commented):
Have you considered that some scenarios have `prepare` playbooks?

molecule/common/molecule/1_vm_not_idempotence.yml
hackallcode: What is better: `not idempotent` or `no idempotence`?

dokshina (dismissed):

dokshina (approved):

PR #322, opened 2021-04-30
Use assert for checking OS

dokshina:
Now `assert` module is used for the `ansible_os_family` check to make logs prettier.

### DIFF
--- a/tasks/prepare.yml
+++ b/tasks/prepare.yml
@@ -3,8 +3,8 @@
 - name: 'Check that OS family is supported'
   assert:
     msg: 'Deploy to {{ ansible_os_family }} distributions is not supported yet'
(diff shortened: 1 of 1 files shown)


hackallcode (dismissed):
Add description and feel free to merge!

tasks/prepare.yml
```diff
@@ -1,11 +1,11 @@
 ---
 
-- name: 'Validate OS Family'
-  fail:
+- name: 'Check that OS family is supported'
+  assert:
     msg: 'Deploy to {{ ansible_os_family }} distributions is not supported yet'
```
hackallcode:
```suggestion
    fail_msg: 'Deploy to {{ ansible_os_family }} distributions is not supported yet'
```

hackallcode (approved):

PR #337, opened 2021-05-28
Fix getting control instance

dokshina:
Fixed fail on getting control instance when there are not unjoined instances with
`replicaset_alias` set.

* If there are joined instances - one alive of them is choosen
* if there are instances in play hosts with `replicaset_alias` set - one alive of them chosen
* otherwise, any alive instance is chosen

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,11 @@ Please update `ansible-galaxy install` command in
 README.md to use the newest tag with new release
 -->
 
+### Fixed
+
+- Fail on getting control instance when there are not unjoined instances with
+  `replicaset_alias` set
+
 ### Added
 
 - `wait_members_alive` step to wait until all cluster members have `alive` status
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ README.md to use the newest tag with new release
 
 ### Fixed
 
-- Fail on getting control instance when there are not unjoined instances with
+- Fail on getting control instance when all unjoined instances haven't
   `replicaset_alias` set
 
 ### Added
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -10,6 +10,11 @@ Please update `ansible-galaxy install` command in
 README.md to use the newest tag with new release
 -->
 
+### Fixed
+
+- Fail on getting control instance when there are not unjoined instances with
+  `replicaset_alias` set
```
hackallcode:
```suggestion
- Fail on getting control instance when there are only unjoined instances without `replicaset_alias`
```

library/cartridge_get_control_instance.py
```diff
@@ -124,19 +125,37 @@ def get_control_instance_name(module_hostvars, play_hosts, control_console):
...
+            return None, "There is no alive joined instances in the cluster"
+    elif to_be_joined_instances:
+        # There are no joined instances, but some unjoined instances
+        # have replicaset_alias set.
+        # One of this instances that is alive should be used for join
+        # and cluster configuration.
+        candidates_uris = to_be_joined_instances.intersection(alive_instances_uris)
+
+        if not candidates_uris:
+            return None, "There is no alive instances that are configured to be joined"
```
hackallcode:
```suggestion
            return None, "There is no alive instances that are should be joined"
```

library/cartridge_get_control_instance.py
```diff
@@ -124,19 +125,37 @@ def get_control_instance_name(module_hostvars, play_hosts, control_console):
...
+        # One of this instances that is alive should be used for join
+        # and cluster configuration.
         candidates_uris = not_joined_instances_uris.intersection(alive_instances_uris)
 
-    if not candidates_uris:
-        return None, "There is no alive instances in the cluster"
+        if not candidates_uris:
+            return None, "There is no alive instances in the cluster"
+
+    assert candidates_uris
```
hackallcode: It's rudiment assert

hackallcode (dismissed):

hackallcode (approved):

PR #341, opened 2021-05-31
Add a step to patch instance parameters in runtime

### LINKED ISSUE #253: [2pt] Add a step to patch instance parameters in runtime
Sometimes we want to change some `box.cfg` parameters in runtime and/or in the configuration files. 
It can be convenient to do it like this:

```yaml
- name: Set replication_connect_quorum to zero
  hosts: "*storage*"
  become: true
  gather_facts: no
  roles:
    - tarantool.cartridge
  vars:
    cartridge_scenario:
      - set_box_cfg_params
    cartridge_box_cfg_params:
      replication_connect_quorum: 0
    runtime_only: true
    # persist_only: true
```

Parameters are set to specified values for all instances specified for the play.

By default, parameters are changed in configuration files.
If some of them can be changed in runtime - it's changed.
In this case, `need_restart` is set to `true` if the instance should be restarted to apply changes.

If `runtime_only` is set, then the step tries to change ALL passed parameters in runtime for each instance.
If `persist_only` is set, all parameters are changed only in the configuration files.

hackallcode:
Closes #253

Sometimes we want to change some `box.cfg` parameters in runtime only.

Now, it's possible to do like this:

```yaml
- name: Set replication_connect_quorum to zero
  hosts: "*storage*"
  become: true
  gather_facts: no
  roles:
    - tarantool.cartridge
  vars:
    cartridge_scenario:
      - patch_instance_in_runtime
    cartridge_runtime_params:
      replication_connect_quorum: 0
```


doc/scenario.md
```diff
@@ -824,3 +825,16 @@ Restart and enable instance systemd service without any conditions.
 Input facts (set by role):
 
 - `instance_info` - information for a current instance ([more details here](#role-facts-descriptions)).
+
+### configure_box_cfg_params
```
dokshina: Can we reflect in a step and variable name the fact that these params are changed only in runtime? 
Maybe `call_box_cfg`? This name says exactly what it does - simply call `box.cfg`.

`configure_instances` step configures instances with patching configuration files, so `configure_box_cfg_params` may confuse - seems that it patches configuration files too.

doc/variables.md
```diff
@@ -78,11 +78,13 @@ For more details see [scenario documentation](/doc/scenario.md).
...
 ## Instances configuration
 
+* `cartridge_defaults` (`dict`, default: `{}`): default configuration
+  parameters values for instances;
 * `config` (`dict`, required): [instance configuration](/doc/instances.md);
 * `zone` (`string`): instance zone (available since
   [Cartridge 2.4.0](https://github.com/tarantool/cartridge/releases/tag/2.4.0));
-* `cartridge_defaults` (`dict`, default: `{}`): default configuration
-  parameters values for instances;
+* `cartridge_box_cfg_params` (`dict``): [instance configuration](/doc/instances.md)
```
dokshina: `box_cfg_params` - parameters to pass into `box.cfg` call? See comment above.

library/cartridge_configure_instance.py
```diff
@@ -0,0 +1,91 @@
...
+    if old_value is None or new_value is None or new_value <= old_value:
+        return False, None
+
+    increase_memory_func_body = '''
+        local param_name, new_memory_size = ...
+        local ok, err = pcall(function()
+            box.cfg {
+                [param_name] = new_memory_size
+            }
+        end)
```
dokshina:
```lua
pcall(box.cfg, {[param_name] = new_memory_size})
```

library/cartridge_configure_instance.py
```diff
@@ -0,0 +1,91 @@
...
+        return False, None
+
+    old_config = helpers.get_box_cfg(control_console)
+    new_config = deepcopy(defaults or {})
+    new_config.update(instance_config)
+
+    return change_dynamic_params(control_console, old_config, new_config)
+
+
+def configure_instance(params):
```
dokshina: What about `configure_instance_dynamic_params`?

This confuses, because of the `configure_instance` step.

module_utils/helpers.py
```diff
@@ -437,6 +437,24 @@ def get_control_console(socket_path):
     return Console(socket_path)
 
 
+def get_control_console_if_started(console_sock):
```
dokshina: Seems that this function shouldn't return an error if it raises.
Or it should return an error only in case when it raises now.

library/cartridge_configure_instance.py
```diff
@@ -0,0 +1,91 @@
...
+
+def change_dynamic_params(control_console, old_config, new_config):
+    memory_changed = False
+    params_to_change = {}
+
+    for param_name, new_value in new_config.items():
+        old_value = old_config.get(param_name)
+
+        if param_name in helpers.MEMORY_SIZE_BOX_CFG_PARAMS:
+            changed, err = change_memory_size(control_console, param_name, old_value, new_value)
```
dokshina: `memory_changed, err = ...`

library/cartridge_configure_instance.py
```diff
@@ -0,0 +1,91 @@
...
+            if err is not None:
+                return None, 'Failed to change memory size in runtime: %s' % err
+            if changed:
+                memory_changed = True
+
+        elif param_name in helpers.DYNAMIC_BOX_CFG_PARAMS and new_value != old_value:
```
dokshina: What if somebody specified non-dynamic param using a new step?

module_utils/helpers.py
```diff
@@ -437,6 +437,24 @@ def get_control_console(socket_path):
...
+
+    try:
+        return get_control_console(console_sock), None
+    except CartridgeException as e:
+        allowed_errcodes = [
+            CartridgeErrorCodes.SOCKET_NOT_FOUND,
+            CartridgeErrorCodes.FAILED_TO_CONNECT_TO_SOCKET,
+            CartridgeErrorCodes.INSTANCE_IS_NOT_STARTED_YET
+        ]
+        if e.code in allowed_errcodes:
```
dokshina: Can `e.code in allowed_errcodes` check be encapsulated in `CartridgeException` class?
To be used like this:

```python
if e.is_instance_not_started_error():
    ...
```

tasks/steps/configure_instance.yml
```diff
@@ -4,9 +4,9 @@
   tags: cartridge-instances
   block:
     - name: 'Manage instance parameters in runtime'
```
dokshina, line 6: Sometimes we could configure `cartridge.cfg` params in runtime, would it be easy to do it with minimal code changes?

library/cartridge_configure_instance.py
hackallcode: It's wrong. If first memory_changed will be true, but second false, then in final it will be false

library/cartridge_configure_instance.py
hackallcode: It will be ignored

tasks/steps/configure_instance.yml
hackallcode: It's impossible to call cartridge.cfg in runtime (only once)

module_utils/helpers.py
hackallcode: I'll remove existing returned errors and change raise to return none, err

module_utils/helpers.py
hackallcode: Ok, I'll do it

library/cartridge_configure_instance.py
hackallcode: configure_instance_in_runtime?

library/cartridge_configure_instance.py
hackallcode: Ok, it's an old code, I'll fix it

doc/scenario.md
hackallcode: patch_instance_in_runtime

doc/variables.md
hackallcode: cartridge_runtime_params

doc/scenario.md
```diff
@@ -824,3 +825,16 @@ Restart and enable instance systemd service without any conditions.
 Input facts (set by role):
 
 - `instance_info` - information for a current instance ([more details here](#role-facts-descriptions)).
+
+### patch_instance_in_runtime
+
+Configure instance in runtime only.
```
dokshina: Patch instance parameters (now only `box.cfg`) in runtime only. If the none-dynamic [link to the list of dynamic `box.cfg` params]  parameter is specified, only dynamic params are changed and error is returned. Note, that memory size can be only increased in runtime.

doc/variables.md
```diff
@@ -78,11 +78,13 @@ For more details see [scenario documentation](/doc/scenario.md).
...
 ## Instances configuration
 
+* `cartridge_defaults` (`dict`, default: `{}`): default configuration
+  parameters values for instances;
 * `config` (`dict`, required): [instance configuration](/doc/instances.md);
 * `zone` (`string`): instance zone (available since
   [Cartridge 2.4.0](https://github.com/tarantool/cartridge/releases/tag/2.4.0));
-* `cartridge_defaults` (`dict`, default: `{}`): default configuration
-  parameters values for instances;
+* `cartridge_runtime_params` (`dict``): [instance configuration](/doc/instances.md)
```
dokshina: Instance configuration parameters to patch in runtime?

library/cartridge_patch_instance_in_runtime.py
```diff
@@ -0,0 +1,108 @@
...
+    incorrect_params_errors = []
+
+    for param_name, new_value in new_config.items():
+        old_box_value = old_box_config.get(param_name)
+
+        if param_name in helpers.MEMORY_SIZE_BOX_CFG_PARAMS:
+            if new_value > old_box_value:
+                memory_sizes_to_change.update({param_name: new_value})
+            else:
+                incorrect_params_errors.append("impossible to change '%s' from '%s' to '%s' in runtime" % (
```
dokshina: Impossible to decrease memory size in runtime.

library/cartridge_patch_instance_in_runtime.py
```diff
@@ -0,0 +1,108 @@
...
+                    param_name, old_box_value, new_value,
+                ))
+
+        elif param_name in helpers.DYNAMIC_BOX_CFG_PARAMS:
+            if new_value != old_box_value:
+                box_params_to_change.update({param_name: new_value})
+
+        else:
+            incorrect_params_errors.append("impossible to change '%s' in runtime" % param_name)
```
dokshina: Maybe `'%s' parameter isn't dynamic and can be changed only via restarting instance`?

BTW, what about one error message for all such parameters? `p_1, p_2, ... parameters aren't dynamic ...`?

module_utils/helpers.py
```diff
@@ -437,6 +444,21 @@ def get_control_console(socket_path):
...
+def get_control_console_if_started(console_sock):
+    if not os.path.exists(console_sock):
+        # Console socket doesn't exists
+        return None, None
+
+    try:
+        return get_control_console(console_sock), None
+    except CartridgeException as e:
+        if e.is_instance_not_started_error():
+            # Impossible to connect to socket
```
dokshina, line 457: All except branch is "Impossible to connect". BTW seems that it should be an error for `strict` mode.

molecule/configure_instance/converge.yml
```diff
@@ -0,0 +1,85 @@
...
+        success_msg: 'Box config params are correct'
+        that:
+          - eval_res[0]['log_level'] == 4
+          - eval_res[0]['memtx_memory'] == 33554433
+
+    ###############
+    # ERROR CASES #
+    ###############
+
+    - name: 'Incorrect update of box memory'
```
dokshina: I think that we need tests with non-dynamic params specified.

molecule/configure_instance/converge.yml
```diff
@@ -0,0 +1,85 @@
...
+      debug:
+        msg: '{{ eval_res }}'
+      run_once: true
+
+    - name: 'Check changed sections'
+      assert:
+        fail_msg: 'Box config params are incorrect'
+        success_msg: 'Box config params are correct'
+        that:
+          - eval_res[0]['log_level'] == 4
```
dokshina: What about storing `new_log_level` and `old_log_level` in variables?

doc/scenario.md
```diff
@@ -824,3 +825,21 @@ Restart and enable instance systemd service without any conditions.
 Input facts (set by role):
 
 - `instance_info` - information for a current instance ([more details here](#role-facts-descriptions)).
+
+### patch_instance_in_runtime
+
+Patch instance parameters (now only box.cfg) in runtime only.
```
dokshina: Maybe "patch dynamic ([link here] with `Dynamic: true`) instance parameters"?

dokshina: box.cfg -> `box.cfg`

library/cartridge_patch_instance_in_runtime.py
```diff
@@ -0,0 +1,115 @@
...
+    if strict_mode:
+        messages = []
+        if incorrect_memory_changes:
+            messages.append('impossible to decrease memory sizes in runtime (%s)' % ', '.join(
+                map(lambda change_info: "'%s' from '%s' to '%s'" % change_info, incorrect_memory_changes)
+            ))
+        if non_dynamic_params:
+            messages.append("impossible to change '%s' in runtime" % ', '.join(non_dynamic_params))
+        if messages:
+            return None, '; '.join(messages)
```
dokshina: The result error message will start with a lowercase letter.

dokshina: What about `"Errors occurred on applying params: %s" % '';".join(messages)` or smth like this?

library/cartridge_patch_instance_in_runtime.py
```diff
@@ -0,0 +1,115 @@
...
+
+def configure_box_cfg_params(console_sock, instance_config, defaults=None, strict_mode=False):
+    control_console, err = helpers.get_control_console_if_started(console_sock, strict_mode)
+    if err is not None:
+        return None, err
+    if not control_console:
+        return False, None
+
+    if not helpers.box_cfg_was_called(control_console):
+        return False, None
```
dokshina, line 95: For strict mode, it should be an error.

doc/scenario.md
```diff
@@ -824,3 +825,22 @@ Restart and enable instance systemd service without any conditions.
 Input facts (set by role):
 
 - `instance_info` - information for a current instance ([more details here](#role-facts-descriptions)).
+
+### patch_instance_in_runtime
+
+Patch dynamic (see [parameters](https://www.tarantool.io/en/doc/latest/reference/configuration/#configuration-parameters)
+with `Dynamic: no`) instance parameters in runtime only
```
dokshina: `Dynamic: yes`.

molecule/patch_instance/converge.yml
```diff
@@ -0,0 +1,98 @@
...
+        msg: '{{ eval_res }}'
+      run_once: true
+
+#    - name: 'Check changed sections'
+#      assert:
+#        fail_msg: 'Box config params are incorrect'
+#        success_msg: 'Box config params are correct'
+#        that:
+#          - eval_res[0]['log_level'] == old_log_level
+#          - eval_res[0]['memtx_memory'] == old_memtx_memory
```
dokshina: Debug?

dokshina (approved):

PR #342, opened 2021-06-04
Fixes for Ansible 4.0

hackallcode:
Task 'Set facts that can be set by the user' fell on Ansible 4.0.

Now it works.
Also added skipped molecule test.


dokshina (approved):
LGTM

PR #343, opened 2021-06-04
Handle bad membership members

dokshina:
Before this patch getting control instance failed on checking membership members
if some member payload is empty. It caused because an empty Lua table is serialized
as an empty array, but the Python module expects a dict. Since this patch, all empty
table cases are handled for membership.


hackallcode (approved):

hackallcode:
Add to PR description problem description

PR #346, opened 2021-06-11
Consider `ansible_port` on determining hosts uniqueness

### LINKED ISSUE #113: [1pt] Determine hosts uniqueness by host + port, not only by host
Now if I setup cluster is such a way that multiple VMs would be accessible through the same hostname but different ssh ports, `ansible-cartridge` would be able to deliver package only to the first VM, not all of them.

That happen because it [filters hosts out](https://github.com/tarantool/ansible-cartridge/blob/c1660e0b7048f6f728debd2cc6af99e8af496f7c/filter_plugins/filter_hosts.py#L28) only by host name, without considering ssh port.

This case might happen when nginx frontend proxy passes tcp ports to other machines or vagrant configuration forwards local ssh ports to VMs.

dokshina:
Closes #113

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ Please update `ansible-galaxy install` command in
 README.md to use the newest tag with new release
 -->
 
+### Changed
+
+- Hosts uniqueness considers `ansible_port`, not only `ansible_host`
+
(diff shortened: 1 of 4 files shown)


hackallcode (approved):
Good patch!

PR #356, opened 2021-06-28
Add availability to set environment variables for instance service

### LINKED ISSUE #331: [4pt] Allow to customize env variable for services
It would be great to have ability to specify env variables per instance. We need it to enable some features on subset of instances.

Example. I want to specify "ENABLE_1_MILLON_RPS=1" and then check it via os.getenv.
```lua
if os.getenv('ENABLE_1_MILLON_RPS') ~= nil then
    box.cfg{}
else
    require('ffi').new('char*')[0]
end

```

hackallcode:
Closes #331

Before the patch, it was impossible to set environment variables for services,
except for some that are used in the unit file template in this role
(so, it was possible only for TGZ).

Now it's possible to set environment variables for all instace services for any type of package.


tasks/steps/configure_instance.yml
```diff
@@ -3,6 +3,10 @@
 - when: not expelled
   tags: cartridge-instances
   block:
+    - name: 'BLOCK: Extend systemd unit files'
```
olegrok, line 6: Why is `BLOCK` in uppercase?

tasks/steps/blocks/extend_systemd_unit_files.yml
```diff
@@ -0,0 +1,31 @@
...
+    path: '{{ (cartridge_systemd_dir, instance_info.systemd_service_dir) | cartridge_path_join }}'
+  register: systemd_service_dir_res
+
+- name: 'Extend systemd unit files with environment'
+  copy:
+    dest: '{{ (cartridge_systemd_dir, instance_info.systemd_service_env_file) | cartridge_path_join }}'
+    content: |
+      [Service]
+      {% for key, value in cartridge_extra_env.items() %}
+      Environment={{ key }}={{ value }}
```
olegrok: Does it work fine if option contains spaces or some special characters? Don't you need some escaping?

tasks/steps/configure_instance.yml
hackallcode: It's a style of role: to show that this task doesn't make any actions, it's start of block

olegrok: Ok. Thanks for explanation.

tasks/steps/blocks/extend_systemd_unit_files.yml
hackallcode: No, but i fixed it and added in tests:
https://github.com/tarantool/ansible-cartridge/blob/3083d473becf787581795f26fa72c8fa057f8dc9/molecule/custom_env/converge.yml#L55-L58

olegrok (dismissed):

defaults/main.yml
```diff
@@ -99,6 +99,7 @@ cartridge_defaults: {}
 config: null
 cartridge_runtime_params: null
 zone: null
+cartridge_extra_env: null
```
dokshina, line 102: `cartridge_instance_env`?

tasks/steps/blocks/extend_systemd_unit_files.yml
```diff
@@ -0,0 +1,31 @@
+---
+
+- name: 'Create directory for systemd unit files extensions'
+  file:
+    state: '{{ "directory" if cartridge_extra_env else "absent" }}'
+    recurse: '{{ true if cartridge_extra_env else false }}'
+    path: '{{ (cartridge_systemd_dir, instance_info.systemd_service_dir) | cartridge_path_join }}'
```
dokshina: It's `systemd_service_dir_name` if you need to join it. BTW, why don't you compute abspath on getting instance info?

tasks/steps/blocks/extend_systemd_unit_files.yml
```diff
@@ -0,0 +1,31 @@
...
+  register: systemd_service_dir_res
+
+- name: 'Extend systemd unit files with environment'
+  copy:
+    dest: '{{ (cartridge_systemd_dir, instance_info.systemd_service_env_file) | cartridge_path_join }}'
```
dokshina: The same as above.

doc/variables.md
```diff
@@ -87,6 +87,8 @@ For more details see [scenario documentation](/doc/scenario.md).
 * `config` (`dict`, required): [instance configuration](/doc/instances.md);
 * `zone` (`string`): instance zone (available since
   [Cartridge 2.4.0](https://github.com/tarantool/cartridge/releases/tag/2.4.0));
+* `cartridge_extra_env` (`dict`): environment variables for instance service
```
dokshina, line 90: What about describing the difference between `null` and `{}`?

defaults/main.yml
hackallcode: You can use this variable for all instances

doc/variables.md
hackallcode: Already....

dokshina (dismissed):

molecule/custom_env/converge.yml
```diff
@@ -0,0 +1,133 @@
...
+
+    - name: 'Get info about service directory'
+      stat:
+        path: '{{ instance_info.systemd_service_dir }}'
+      register: stat_res
+
+    - name: 'Check service directory'
+      assert:
+        fail_msg: "Directory should exist"
+        success_msg: "Directory shouldn't exist"
```
dokshina, line 132:
```
        fail_msg: "Directory exist"
        success_msg: "Directory shouldn't exist"
```

molecule/custom_env/converge.yml
```diff
@@ -0,0 +1,133 @@
...
+        fail_msg: 'Received bad environment'
+        success_msg: 'Received environment is OK'
+        that: eval_res == []
+
+    - name: 'Get info about service directory'
+      stat:
+        path: '{{ instance_info.systemd_service_dir }}'
+      register: stat_res
+
+    - name: 'Check service directory'
```
dokshina: `Check service directory doesn't exist`

molecule/custom_env/converge.yml
```diff
@@ -0,0 +1,133 @@
...
+
+    - name: 'Get info about service environment file'
+      stat:
+        path: '{{ instance_info.systemd_service_env_file }}'
+      register: stat_res
+
+    - name: 'Check service environment file'
+      assert:
+        fail_msg: "Environment file shouldn't exist"
+        success_msg: "Environment file should exist"
```
dokshina, line 90:
```
        fail_msg: "Environment file doesn't exist"
        success_msg: "Environment file should exist"
```

molecule/custom_env/converge.yml
```diff
@@ -0,0 +1,133 @@
...
+
+    - name: 'Get info about service directory'
+      stat:
+        path: '{{ instance_info.systemd_service_dir }}'
+      register: stat_res
+
+    - name: 'Check service directory'
+      assert:
+        fail_msg: "Directory should exist"
+        success_msg: "Directory shouldn't exist"
```
dokshina, line 41:
```
        fail_msg: "Directory exists"
        success_msg: "Directory shouldn't exist"
```

dokshina (commented):
Maybe I've missed other sich places, please check

dokshina (approved):

PR #364, opened 2021-07-05
Add step to disable/enable instances

### LINKED ISSUE #340: [8pt] Add step to disable/enable instances
It may be needed when some instances from replicasets were dropped (for example leader in storage replicaset) and you can't promote a new leader, because you can't apply two-phase commit.

hackallcode:
Closes #340

Sometimes you can't promote a new leader,
because you have broken instances and can't apply two-phase commit.

Now you can disable broken instances by `disabled` flag in instance config.

Since the disabled instance doesn't receive topology updates, it doesn't know that it's disabled.
So, we had to do the following to detect what instances are disabled (e.g., it's necessary to select control instance):

1. Ignore stateboards, non-running instances, and instances without a topology section (they have no information about disabled instances);
2. Collect from each instance checksum of the topology section and the list of disabled instances;
3. Determine for each instance config mismatches with the enabled (according to its version) instances;
4. If the first instance has a config mismatch with the second instance, but the second instance thinks that everything is OK, it means that the second instance didn't receive topology updates and it was disabled, so we ignore it;
5. On instances with correct configs, we collect information about disabled servers;
6. If there are several versions of disabled instances, then we raise an error that we may have a split-brain. If the `cartridge_ignore_split_brain` flag is set, we ignore this error and take the version with more voices.

### DIFF
--- a/doc/steps.md
+++ b/doc/steps.md
@@ -68,6 +68,9 @@ Some of useful variables always establishes during role preparation, so them can
   - `paths_to_remove_on_expel` - paths that will be removed on instance expel;
   - `files_to_remove_on_cleanup` - files that will be removed on instance cleanup;
   - `dirs_to_remove_on_cleanup` - dirs that will be removed on instance cleanup;
+  - `cluster_disabled_instances` - list of disabled instances from instance config;
+  - `cluster_topology_checksum` - checksum of topology section in instance config;
+- `cluster_disabled_instances` - disabled instances in cluster config;
 - `single_instances_for_each_machine` - list of instances (Ansible hosts), one for each physical machine,
   for example, can be used in `delegate_to`;
 - `instances_from_same_machine` - dictionary, where key is the hostname of the instance,
@@ -139,8 +142,10 @@ Input variables from config:
 
 - `cartridge_package_path` - should be specified to compute app distribution directory
   (otherwise, `update_instance` is skipped);
-- `expelled` - indicates if instance must be expelled from topology;
-- `restarted` - if instance should be restarted or not (user forced decision);
+- `expelled` - indicates if instance must be expelled from topology
+  (to determine if it's should be checked if instance should be restarted);
+- `restarted` - if instance should be restarted or not (user forced decision)
+  (to determine if it's should be checked if instance should be restarted);
 - `cartridge_multiversion` - indicates that
   [multiversion approach](/doc/multiversion.md) is enabled;
 - `cartridge_app_user` - user which will own the links;
@@ -259,6 +264,7 @@ Edit topology of replicasets.
 Input variables from config (basic):
 
 - `expelled` - indicates if instance must be expelled from topology;
+- `disabled` - indicates if instance should be disabled;
 - `stateboard` - indicates that the instance is a stateboard;
 - `replicaset_alias` - replicaset alias, will be displayed in Web UI;
 - `roles` - roles to be enabled on the replicaset;
@@ -341,6 +347,7 @@ Bootstrap VShard in cluster.
 Input variables from config:
 
 - `expelled` - indicates if instance must be expelled from topology;
+- `disabled` - indicates if instance should be disabled;
 - `stateboard` - indicates that the instance is a stateboard;
 - `cartridge_bootstrap_vshard` - indicates if vshard should be bootstrapped;
 - `bootstrap_vshard_retries` - retries to bootstrap vshard;
--- a/doc/steps.md
+++ b/doc/steps.md
@@ -68,9 +68,10 @@ Some of useful variables always establishes during role preparation, so them can
   - `paths_to_remove_on_expel` - paths that will be removed on instance expel;
   - `files_to_remove_on_cleanup` - files that will be removed on instance cleanup;
   - `dirs_to_remove_on_cleanup` - dirs that will be removed on instance cleanup;
-  - `cluster_disabled_instances` - list of disabled instances from instance config;
-  - `cluster_topology_checksum` - checksum of topology section in instance config;
-- `cluster_disabled_instances` - disabled instances in cluster config;
+  - `disabled_instances` - list of disabled instances from instance config;
+  - `topology_checksum` - checksum of topology section from instance config;
+- `cluster_disabled_instances` - list of disabled instances from cluster config;
+- `inventory_disabled_instances` - list of disabled instances from inventory;
 - `single_instances_for_each_machine` - list of instances (Ansible hosts), one for each physical machine,
   for example, can be used in `delegate_to`;
 - `instances_from_same_machine` - dictionary, where key is the hostname of the instance,
--- a/doc/variables.md
+++ b/doc/variables.md
@@ -95,6 +95,7 @@ For more details see [scenario documentation](/doc/scenario.md).
   restarted or not (if this flag isn't specified, instance will be restarted if
   it's needed to apply configuration changes);
 * `expelled` (`boolean`, default: `false`): a boolean flag that indicates if instance must be expelled from topology;
+* `disabled` (`boolean`, default: `false`): a boolean flag that indicates if instance should be disabled;
 * `stateboard` (`boolean`, default: `false`): a boolean flag that indicates
    that the instance is a [stateboard](/doc/stateboard.md);
 * `instance_start_retries` (`number`, default: 10) retries to check that all instances become started;
--- a/doc/variables.md
+++ b/doc/variables.md
@@ -8,11 +8,14 @@ vshard bootstrapping, and failover.
 * `cartridge_app_name` (`string`): application name, required;
 * `cartridge_cluster_cookie` (`string`): cluster cookie for all
   cluster instances;
-* `cartridge_not_save_cookie_in_app_config` (`boolean`, default: `false`) - flag indicates that cluster cookie shouldn't be persisted in application configuration file;
-* `cartridge_remove_temporary_files` (`boolean`, optional, default: `false`):
+* `cartridge_not_save_cookie_in_app_config` (`boolean`, default: `false`): flag
+  indicates that cluster cookie shouldn't be persisted in application configuration file;
+* `cartridge_remove_temporary_files` (`boolean`, optional, default: `false`): flag
   indicates if temporary files should be removed
   (more details in description of [`cleanup` step API](/doc/steps.md#step-cleanup));
-* `cartridge_paths_to_keep_on_cleanup` (`list-of-strings`, optional, default: `[]`) -
+* `cartridge_ignore_split_brain` (`boolean`, optional, default: `false`): flag
+  indicates that detected split brain should be ignored on preparation stage;
+* `cartridge_paths_to_keep_on_cleanup` (`list-of-strings`, optional, default: `[]`):
   list of full paths or relative paths to work/memtx/vinyl/wal
   directory that should be kept on instance cleanup
   (it's possible to use bash patterns, e.g. `*.control`).
--- a/library/cartridge_edit_topology.py
+++ b/library/cartridge_edit_topology.py
@@ -68,6 +68,8 @@ def get_all_new_instances(module_hostvars):
         if helpers.is_expelled(instance_vars):
             instance['expelled'] = True
         else:
+            if instance_vars.get('disabled') is not None:
+                instance['disabled'] = instance_vars['disabled']
             if instance_vars.get('zone') is not None:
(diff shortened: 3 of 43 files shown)


hackallcode (commented):
Self-reviewed, all tests are passed: https://github.com/tarantool/ansible-cartridge/actions/runs/1007314200

hackallcode (commented):
Self-reviewed, all tests passed: https://github.com/tarantool/ansible-cartridge/actions/runs/1015459845

doc/steps.md
```diff
@@ -68,6 +68,9 @@ Some of useful variables always establishes during role preparation, so them can
   - `paths_to_remove_on_expel` - paths that will be removed on instance expel;
   - `files_to_remove_on_cleanup` - files that will be removed on instance cleanup;
   - `dirs_to_remove_on_cleanup` - dirs that will be removed on instance cleanup;
+  - `cluster_disabled_instances` - list of disabled instances from instance config;
+  - `cluster_topology_checksum` - checksum of topology section in instance config;
+- `cluster_disabled_instances` - disabled instances in cluster config;
```
dokshina: currently disabled instances?

doc/variables.md
```diff
@@ -95,6 +95,7 @@ For more details see [scenario documentation](/doc/scenario.md).
   restarted or not (if this flag isn't specified, instance will be restarted if
   it's needed to apply configuration changes);
 * `expelled` (`boolean`, default: `false`): a boolean flag that indicates if instance must be expelled from topology;
+* `disabled` (`boolean`, default: `false`): a boolean flag that indicates if instance should be disabled;
```
dokshina, line 101: But `disabled: false` means that the instance should be enabled if it was disabled.

library/cartridge_get_control_instance.py
```diff
@@ -84,17 +85,16 @@ def get_twophase_commit_versions(control_console, advertise_uris):
     return versions, err
 
 
-def candidate_is_ok(uri, names_by_uris, module_hostvars):
+def candidate_is_ok(uri, names_by_uris, module_hostvars, cluster_disabled_instances):
     instance_name = names_by_uris[uri]
-    instance_vars = module_hostvars[instance_name]
-
-    if helpers.is_expelled(instance_vars):
+    if instance_name in cluster_disabled_instances:
```
dokshina, line 90: `cluster_disabled_instances` name confuses when used before `helpers.is_enabled`. Names don't reflect that some instances are currently disabled, and others should be disabled.

library/cartridge_get_disabled_instances.py
```diff
@@ -0,0 +1,78 @@
...
+
+from ansible.module_utils.helpers import Helpers as helpers
+
+argument_spec = {
+    'module_hostvars': {'required': True, 'type': 'dict'},
+    'play_hosts': {'required': True, 'type': 'list'},
+}
+
+
+def get_disabled_instances(instance_vars):
```
dokshina: This function looks like it performs the main module action. I can't imagine a better name, it's smth like "instances disabled for one specific instance"?

library/cartridge_get_disabled_instances.py
```diff
@@ -0,0 +1,78 @@
...
+            get_topology_checksum(module_hostvars[name]) is not None,
+        ]),
+        play_hosts,
+    ))
+
+    for instance_name in play_hosts:
+        disabled_instances = get_disabled_instances(module_hostvars[instance_name])
+
+        not_disabled_names = list(filter(lambda other_name: other_name not in disabled_instances, play_hosts))
+        if check_config_mismatch(module_hostvars, instance_name, not_disabled_names):
```
dokshina: `if config_mismatched`?

library/cartridge_get_disabled_instances.py
```diff
@@ -0,0 +1,78 @@
...
+            config_mismatch_count += 1
+            continue
+
+        started_count += 1
```
dokshina: Generally, an instance can be both`started` and `config_mismatch`. 
In this case, this check can be

```python
# started_count == 0 and config_mismatch_count > 0
config_mismatch_count >= started_count
```

library/cartridge_get_disabled_instances.py
```diff
@@ -0,0 +1,78 @@
...
+        for disabled_instance in disabled_instances:
+            rating[disabled_instance] = rating.get(disabled_instance, 0) + 1
+
+    if started_count == 0 and config_mismatch_count > 0:
+        return helpers.ModuleRes(failed=True, msg='All instances in cluster has different topology configs')
+
+    final_disabled_instances = []
+    split_brain_exists = False
```
dokshina: `split_brain_detected`? `split_brain_suspected`?

library/cartridge_get_disabled_instances.py
```diff
@@ -0,0 +1,78 @@
...
+
+    for name, score in rating.items():
```
dokshina: Name `rating` says nothing. `votes_to_disable`?

library/cartridge_get_disabled_instances.py
```diff
@@ -0,0 +1,78 @@
...
+        if score >= started_count / 2:
+            final_disabled_instances.append(name)
+        if score != started_count:
+            split_brain_exists = True
+
+    if split_brain_exists:
+        helpers.warn("It seems that you have split brain in your cluster")
+
+    return helpers.ModuleRes(changed=False, fact=sorted(final_disabled_instances))
```
dokshina: Maybe we need a possibility to get a list of instances to be disabled before running `edit_topology`? Is this fact described in the doc? This operation seems dangerous.

library/cartridge_get_facts_for_machines.py
```diff
@@ -41,7 +43,8 @@ def get_facts_for_machines(params):
 
         # Calculate single not expelled instance for each machine
         if all([
-            not helpers.is_expelled(instance_vars),
+            helpers.is_enabled(instance_vars),
+            instance_name not in cluster_disabled_instances,
```
dokshina, line 47: What about `helpers.is_enabled(instance_vars, cluster_disabled_instances)`? I've seen this code before.

module_utils/helpers.py
```diff
@@ -610,6 +619,48 @@ def patch_clusterwide_config(control_console, new_sections):
     return True, None
 
 
+def get_disabled_instances(console_sock, stateboard):
```
dokshina, line 638: `get_currently_disabled_instances`?

module_utils/helpers.py
```diff
@@ -610,6 +619,48 @@ def patch_clusterwide_config(control_console, new_sections):
...
+    return control_console.eval_res_err('''
+        local digest = require('digest')
+        local confapplier = require('cartridge.confapplier')
+
+        local active_config = confapplier.get_active_config()
+        if active_config == nil then
+            return nil
+        end
+
+        local topology_section = active_config:get_plaintext('topology.yml')
```
dokshina, line 671: Are we support versions of Cartridge that have another config format?

dokshina (commented):
I am sure that disabled instances should be added to `molecule/dead_instance`, and `molecule/default` tests.

doc/variables.md
hackallcode: Yes, it works as `expelled` flag

library/cartridge_get_control_instance.py
hackallcode: helpers.is_enabled works only with inventory, we can make helpers.is_enabled_in_inventory

library/cartridge_get_disabled_instances.py
hackallcode: started_count == healthy_count. So, it's possible that 1 healthy and 3 with config mismatch. So, we should use information only from 1 healthy instance (if 3 others disabled)

library/cartridge_get_disabled_instances.py
hackallcode: But this fact is set at the preparation stage...

hackallcode: Maybe it's better to update list after edit topology

library/cartridge_get_facts_for_machines.py
hackallcode: We should pass instance_name too. So, it will be strange: `helpers.is_enabled(instance_vars, cluster_disabled_instances, instance_name)`

library/cartridge_get_disabled_instances.py
hackallcode: I made two lists: inventory_disabled_instances and expected_cluster_disabled_instances

doc/variables.md
dokshina: No, you can't undo expelling instance. 

expelled indicates that instance should be expelled;
disabled indicates that instance should be disabled or enabled.

module_utils/helpers.py
hackallcode: Yes, it works with Cartridge 2.1.2

library/cartridge_get_disabled_instances.py
```diff
@@ -0,0 +1,95 @@
...
+    split_brain_suspected = False
+
+    for name, score in votes_to_disable.items():
+        if score >= healthy_count / 2:
+            final_disabled_instances.append(name)
+        if score != healthy_count:
+            split_brain_suspected = True
+
+    if split_brain_suspected:
+        helpers.warn("It seems that you have split brain in your cluster")
```
Kasen: I think we should stop playbook execution here, cause split brain should be fixed by man.

Kasen (dismissed):
LGTM

library/cartridge_get_disabled_instances.py
hackallcode: Now, it's an error, but you can ignore it by `cartridge_ignore_split_brain` flag

dokshina:
I have some suggestions about the PR description:

> Since the cartridge does not allow the instance to know if it is disabled, we had to do the following:

Since the disabled instance doesn't receive topology updates, it doesn't know that it's disabled.

> Since the cartridge does not allow the instance to know if it is disabled, we had to do the following:

When do we do the following? Please describe that all these things are needed to choose a control instance.

> If someone has a config mismatch, but another instance does not have a config mismatch, then this means that the instance is not in the cluster (disabled, expelled, ...) and we ignore it;

It's hard to understand. Which instance is not in the cluster? 
If the first instance has a config mismatch with the second instance, but the second instance thinks that everything is OK, it means that second instance didn't receive topology updates and it was disabled, so we don't choose it as a control instance.

> On instances with correct configs, collect statistics about disabled servers;

What statistics do you mean?

library/cartridge_edit_topology.py
```diff
@@ -489,6 +499,9 @@ def get_server_params(instance_name, new_instance, old_instances, allow_missed_i
     if new_instance.get('expelled') is True:
         server_params['expelled'] = True
     else:
+        for flag_name in ['disabled']:
```
dokshina, line 502: There is no need to add other flags for such a server, let's continue.

library/cartridge_get_disabled_instances.py
```diff
@@ -0,0 +1,102 @@
...
+    final_disabled_instances = []
+    split_brain_suspected = False
+
+    for name, score in votes_to_disable.items():
+        if score >= healthy_count / 2:
+            final_disabled_instances.append(name)
+        if score != healthy_count:
+            split_brain_suspected = True
+
+    if split_brain_suspected:
```
dokshina: It's not `suspected`, it's `detected`. `suspected` means that you need additional checks to be sure it's true.

tasks/steps/wait_members_alive.yml
```diff
@@ -18,6 +18,5 @@
       until: not check_members_alive_res.failed
       retries: '{{ wait_members_alive_retries }}'
       delay: '{{ wait_members_alive_delay }}'
-      when: not expelled
       run_once: true
```
dokshina, line 21: In fact, it's a bugfix that should be committed separately =)

library/cartridge_edit_topology.py
hackallcode: No, this is done to ensure that the instance is always up to date. So, if suddenly it is turned on through the UI, nothing would break

tasks/steps/wait_members_alive.yml
hackallcode: But in this PR we make the disabled flags near by expelled flags, so we can say that this is part of the PR :)

dokshina (approved):
Oh, bless you!

PR #365, opened 2021-07-06
Fix rolling update doc

FizikRoot:
Before this patch, there was an incorrect example of a leader change in the rolling update document.

Now it has been fixed.


hackallcode (approved):
Thanks for the patch, we didn't notice this typo.

PR #370, opened 2021-07-14
Add step to compare inventory and real cluster topology

### LINKED ISSUE #215: [8pt] Add step to compare inventory and real cluster topology
In real production cases when we have huge size cluster we use any kind of inventory generators. Also we use this generator when we want to change some settings, for example memtx_memory for all storages. But what if we mistaken in code of inventory generator, for example new inventory has another sequence of advertise_uri ports for instances? In such a situation if we run playbook we will get a brick instead of working cluster.

It wiil be great if we will add a step which can spawn current topology from working cluster, compare it with inventory and if all dangerous changes check passes only then we will continue to play. Of course, we need to provide a flag for override check or a way to diable checks.

### LINKED ISSUE #362: [2pt] edit_topology fails if at least one replicaset alias in runnig cluster is not equal to those replicaset alias in ansible inventory
Steps to reproduce:

1. Rollout cluster with some inventory
2. Change replicaset alias in Cartridge UI
3. Run any playbook with edit_topology step

After that you will face the following error: 
Failed to edit topology: servers[xxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxx].uri "advertise_uri:port" collision with another server

hackallcode:
Closes #215
Closes #362

Before the patch, it was dangerous to edit topology.

Now, in `edit_topology` step has been added the check, which compares the current
topology from the working cluster with inventory and if all dangerous changes check
passes only then it will continue to play.
It's possible to call this check by the `check_new_topology` step.
Also, it's possible to disable checks by flags.


CHANGELOG.md
```diff
@@ -14,7 +14,8 @@ README.md to use the newest tag with new release
 
 - Step `cleanup_instance_files` to clean up data of stopped instance;
 - Add availability to set environment variables for instance service;
-- Add `instances_from_same_machine` variable in preparation.
+- Add `instances_from_same_machine` variable in preparation;
+- Add `edit_topology_check` step to compare inventory and real cluster topology.
```
dokshina: I don't think it's a good idea to keep `.` after last line and change two lines on each PR.

doc/steps.md
```diff
@@ -40,6 +40,7 @@ but can be used in a custom one:
 - [restart_instance_force](#step-restart_instance_force)
 - [patch_instance_in_runtime](#step-patch_instance_in_runtime)
 - [cleanup_instance_files](#step-cleanup_instance_files)
+- [edit_topology_check](#step-edit_topology_check)
```
dokshina: Maybe `check_new_topology`?

doc/steps.md
```diff
@@ -509,3 +518,35 @@ Input variables from config:
 - `cartridge_paths_to_keep_on_cleanup` - list of full paths or relative paths
   to work/memtx/vinyl/wal directory that should be kept on instance cleanup
   (it's possible to use bash patterns, e.g. `*.control`).
+
+## Step `edit_topology_check`
+
+Check for dangerous changes on edit topology
+(runs automatically before [edit_topology](#step-edit_topology) step).
```
dokshina: This step runs or this check is performed?

doc/steps.md
```diff
@@ -509,3 +518,35 @@ Input variables from config:
...
+  while two-phase commit (Cartridge 2.5+ is required);
+- `twophase_apply_config_timeout` - time in seconds to wait config apply
+  while two-phase commit (Cartridge 2.5+ is required);
+- `cartridge_force_advertise_uris_change` - flag that disable check
+  for advertise uris change;
+- `cartridge_ignore_extra_cluster_instances` - flag that disable check
+  for instances from the cluster that are not in inventory;
+- `cartridge_ignore_extra_cluster_replicasets` - flag that disable check
+  for replicasets from the cluster that are not in inventory;
+- `cartridge_ignore_renamed_replicasets` - flag that disable check
```
dokshina: Such a big list... Maybe it's better to say "All `edit_topology` facts and additionally..."?

library/cartridge_edit_topology.py
```diff
@@ -125,6 +151,11 @@ def set_enabled_roles(replicasets, control_console):
         })
 
 
+#############################################################
+# Collect difference between real and inventory replicasets #
```
dokshina: What about `real and configured`?

library/cartridge_edit_topology.py
```diff
@@ -44,12 +51,48 @@
...
+    return instances
+
+
+def get_instances_to_configure(module_hostvars, play_hosts):
+    module_hostvars = {
+        instance_name: instance_vars
+        for instance_name, instance_vars in module_hostvars.items()
+        if instance_name in play_hosts
+    }
+    return get_all_instances_from_inventory(module_hostvars)
```
dokshina: `get_all_instances_from_inventory` -> `get_all_instances_from_hostvars`? Generally nventory is all hostvars.

library/cartridge_edit_topology.py
```diff
@@ -347,6 +386,11 @@ def get_replicasets_params_for_changing_failover_priority(replicasets, cluster_r
     return replicasets_params, None
 
 
+#########################################################
+# Collect difference between real and inventory servers #
```
dokshina: `servers` -> `instances`. An inventory does not have configured servers. Thi name is used only because of `edit_topology` params names.

dokshina: BTW, I don't like the word `real`. Maybe `current`?

library/cartridge_edit_topology.py
```diff
@@ -407,51 +454,80 @@ def get_servers_params(instances, cluster_instances, allow_missed_instances):
...
-    )
-    if err is not None:
-        return None, err
+    for cluster_replicaset_name, cluster_replicaset_params in cluster_replicasets.items():
+        if cluster_replicaset_name not in all_replicasets:
+            renamed = False
 
-    if replicasets_params:
-        topology_params['replicasets'] = replicasets_params
+            for replicaset_name, replicaset_params in all_replicasets.items():
```
dokshina: I suggest checking `replicaset_alias` for each instance:

* `cluster_instance[replicaset_alias] == instance[replicaset_alias]` - OK
* `cluster_instance[replicaset_alias] == nil` - OK
* `cluster_instance[replicaset_alias] != instance[replicaset_alias]` - BAD: replicaset was renamed or instance was moved to the other one.

Generally, I'm not sure we should check it. Currently, replicaset renaming isn't supported since the alias is a replicaset identifier. It is not a "dangerous" change, it's impossible. So, it can be non-skipped error on `edit_topology` while we don't allow to rename replicasets.

library/cartridge_edit_topology.py
```diff
@@ -407,51 +454,80 @@ def get_servers_params(instances, cluster_instances, allow_missed_instances):
...
+def check_for_dangerous_changes(
+    all_instances, instances, cluster_instances,
+    all_replicasets, replicasets, cluster_replicasets,
+    ignore_checks,
+):
+    errors = []
+    errors += check_instances_for_dangerous_changes(all_instances, instances, cluster_instances, ignore_checks)
+    errors += check_replicasets_for_dangerous_changes(all_replicasets, replicasets, cluster_replicasets, ignore_checks)
+    if errors:
+        return '. '.join(errors)
```
dokshina: Maybe `"Found dangerous changes: %s. To ignore some of them use these flags..."`

molecule/edit_topology_check/converge.yml
```diff
@@ -0,0 +1,166 @@
...
+- name: 'Rename core_1_replicaset replicaset'
+  hosts: core_1_replicaset
+  become: true
+  become_user: root
+  gather_facts: false
+  tasks:
+    - name: 'Set new alias'
+      set_fact:
+        replicaset_alias: core-rs-1-renamed
+    - debug: var=replicaset_alias
```
dokshina: Debug.

molecule/edit_topology_check/converge.yml
```diff
@@ -0,0 +1,166 @@
...
+
+    - name: 'Show check result'
+      debug:
+        var: edit_topology_check_res
+
+    - name: 'Check that edit topology check message is correct'
+      assert:
+        that: >-
+          edit_topology_check_res.msg == [
+            'Some instances from cluster are missing in inventory: core-2, core-2-replica',
```
dokshina: `are missed`?

doc/steps.md
hackallcode: Tasks from this step runs

library/cartridge_edit_topology.py
hackallcode: old and new

library/cartridge_edit_topology.py
hackallcode: No, inventory is a common term, hostvars is the name of a variable

library/cartridge_edit_topology.py
hackallcode: We haven't replicaset_alias for cluster instances

hackallcode: Also, there are many situations when checking a replica set through INSTANCE will give false-positive results

molecule/edit_topology_check/converge.yml
hackallcode: No, https://dictionary.cambridge.org/dictionary/english-russian/missing

doc/steps.md
```diff
@@ -274,6 +278,17 @@ Input variables from config:
 - `edit_topology_healthy_timeout` - time in seconds to wait until a cluster become healthy after editing topology;
 - [DEPRECATED] `edit_topology_timeout` - the same timeout as `edit_topology_healthy_timeout`.
 
+Input variables from config (for dangerous checks):
```
dokshina: Maybe simply `for new topology check`?

doc/variables.md
```diff
@@ -166,6 +166,17 @@ For more details see [scenario documentation](/doc/scenario.md).
 * `cartridge_eval_retries` number of eval retries;
 * `cartridge_eval_delay` - eval retries delay.
 
+## Flags to disable edit topology checks
```
dokshina: `to ignore`?

library/cartridge_edit_topology.py
```diff
@@ -44,12 +51,48 @@
 ''' % (helpers.FORMAT_REPLICASET_FUNC, helpers.FORMAT_SERVER_FUNC)
 
 
-def get_configured_replicasets(module_hostvars, play_hosts):
-    replicasets = {}
+#########################################
+# Collect information about new cluster #
```
dokshina: `new topology`? It's `check_new_topology`.

library/cartridge_edit_topology.py
```diff
@@ -407,51 +454,96 @@ def get_servers_params(instances, cluster_instances, allow_missed_instances):
...
+        elif instance_name in new_instances:
+            new_uri = new_instances[instance_name].get('uri')
+            old_uri = old_instances[instance_name].get('uri')
+            if new_uri != old_uri:
+                changed_advertise_uris.append('%s (%s -> %s)' % (instance_name, old_uri, new_uri))
 
-    servers_params, err = get_servers_params(instances, cluster_instances, allow_missed_instances)
-    if err is not None:
-        return None, err
+    if extra_cluster_instances and not ignore_checks.get(EXTRA_CLUSTER_INSTANCES_CHECK_NAME):
```
dokshina: What do you think about adding a warning if a check is ignored?

library/cartridge_edit_topology.py
```diff
@@ -407,51 +454,96 @@ def get_servers_params(instances, cluster_instances, allow_missed_instances):
...
-    if err is not None:
-        return None, err
+    for old_replicaset_name, old_replicaset in old_replicasets.items():
+        if old_replicaset_name not in all_new_replicasets:
+            renamed = False
 
-    if replicasets_params:
-        topology_params['replicasets'] = replicasets_params
+            for new_replicaset_name, new_replicaset in all_new_replicasets.items():
+                if set(new_replicaset.get('instances', [])) == set(old_replicaset.get('instances', [])):
```
dokshina: `if ... != ...: continue`?

library/cartridge_edit_topology.py
```diff
@@ -407,51 +454,96 @@ def get_servers_params(instances, cluster_instances, allow_missed_instances):
...
+    extra_cluster_replicasets = []
+    renamed_replicasets = []
 
-    replicasets_params, err = get_replicasets_params_for_changing_failover_priority(
-        replicasets, cluster_replicasets, cluster_instances, allow_missed_instances
-    )
-    if err is not None:
-        return None, err
+    for old_replicaset_name, old_replicaset in old_replicasets.items():
+        if old_replicaset_name not in all_new_replicasets:
```
dokshina: `if .. in ...: continue`? Too many levels.

library/cartridge_edit_topology.py
```diff
@@ -407,51 +454,96 @@ def get_servers_params(instances, cluster_instances, allow_missed_instances):
...
+            if not renamed:
+                extra_cluster_replicasets.append(old_replicaset_name)
 
-    return topology_params, None
+    if extra_cluster_replicasets and not ignore_checks.get(EXTRA_CLUSTER_REPLICASETS_CHECK_NAME):
+        msg = 'some replicasets from cluster are missing in inventory: %s' % ', '.join(extra_cluster_replicasets)
+        errors.append(msg)
+
+    if renamed_replicasets and not ignore_checks.get(RENAMED_REPLICASETS_CHECK_NAME):
+        msg = 'looks like that some replicasets should be renamed in inventory: %s' % ', '.join(renamed_replicasets)
```
dokshina: `should be renamed` -> `has different names`?

library/cartridge_edit_topology.py
```diff
@@ -519,75 +716,34 @@ def edit_topology(params):
     # * Configure instances that are already joined.
     #   New instances aren't configured here since they don't have
     #   UUIDs before join.
-    topology_params, err = get_topology_params(
-        replicasets, cluster_replicasets, instances, cluster_instances, allow_missed_instances
+
+    first_call_changed, err = single_edit_topology_call(
```
dokshina: `first_call_changed` -> `changed_on_first_call`

molecule/check_new_topology/converge.yml
```diff
@@ -0,0 +1,194 @@
...
+          advertise_uri: 'vm1:3305'
+          http_port: 8105
+
+    - vars:
+        cartridge_force_advertise_uris_change: true
+        cartridge_scenario:
+          - configure_instance
+          - restart_instance
+          - wait_instance_started
+          - connect_to_membership
```
dokshina, line 21: Is this step really needed there? There are two versions of one member after connecting to membership?

molecule/check_new_topology/converge.yml
```diff
@@ -0,0 +1,194 @@
...
+          - edit_topology
+      import_role:
+        name: ansible-cartridge
+
+- name: 'Set default scenario'
```
dokshina: What about adding a custom scenario instead of this?

molecule/check_new_topology/converge.yml
```diff
@@ -0,0 +1,194 @@
...
+        var: edit_topology_check_res
+
+    - name: 'Check that edit topology check message is correct'
+      assert:
+        that: >-
+          edit_topology_check_res.msg == (
+            "Found dangerous changes: "
+            "some instances from cluster are missing in inventory: core-2, core-2-replica; "
+            "some replicasets from cluster are missing in inventory: core-rs-2; "
+            "looks like that some replicasets should be renamed in inventory: core-rs-1-renamed -> core-rs-1. "
```
dokshina: It's not clear which name is new, for advertise URIs `(old -> new)` is shown.

molecule/check_new_topology/converge.yml
```diff
@@ -0,0 +1,194 @@
...
+  gather_facts: false
+  tasks:
+    - name: 'Remove instance specific facts'
+      set_fact:
+        config: null
+        replicaset_alias: null
+        failover_priority: null
+        roles: null
+
+- name: 'Rename core_1_replicaset replicaset'
```
dokshina: `... in inventory`

The same for others.

library/cartridge_edit_topology.py
hackallcode: No, it's a hint for user

molecule/check_new_topology/converge.yml
hackallcode: Yes, after this membership will have 5 instances instead of 4

hackallcode: I add this info in name of task

molecule/check_new_topology/converge.yml
hackallcode: Yes

library/cartridge_edit_topology.py
```diff
@@ -407,51 +454,122 @@ def get_servers_params(instances, cluster_instances, allow_missed_instances):
...
+        msg %= ', '.join(extra_cluster_instances)
+        if ignore_errors_of_checks.get(EXTRA_CLUSTER_INSTANCES_CHECK_NAME):
+            helpers.warn(msg)
+        else:
+            errors.append(msg)
 
-    if replicasets_params:
-        topology_params['replicasets'] = replicasets_params
+    if changed_advertise_uris:
+        msg = 'advertise uris of some instances will be changed, you should check it: %s'
```
dokshina: Maybe 

`Advertise uris of some instances were changed: ... . To apply this changes, set advertise_uris_changed flag`

?

molecule/check_new_topology/converge.yml
```diff
@@ -0,0 +1,203 @@
...
+      debug:
+        var: edit_topology_check_res
+      run_once: true
+
+    - name: 'Check that edit topology check message is correct'
+      assert:
+        that: >-
+          edit_topology_check_res.msg == (
+            "Found dangerous changes in topology: "
+              "some instances from cluster are missing in inventory, you should add it: "
```
dokshina: add them.

molecule/check_new_topology/converge.yml
```diff
@@ -0,0 +1,203 @@
...
+                "core-2, core-2-replica; "
+              "advertise uris of some instances will be changed, you should check it: "
+                "core-1-replica (vm1:3305 -> vm1:3306); "
+              "some replicasets from cluster are missing in inventory, you should add it: "
```
dokshina: add them.

dokshina (approved):

PR #380, opened 2021-11-08
Optimize `Set instance facts` step

### LINKED ISSUE #385: [2sp] Optimize Set instance facts step
On the second role import, the `Set instance facts` step takes more time than the first time.

hackallcode:
Closes #385

Before the patch, the `Set instance facts` step used temporary role facts,
although this is not necessary. Therefore, in large clusters, it might take longer than necessary.

Now at this step, only those facts that can be specified by the user are established,
which speeds up the step.


dokshina (dismissed):
Brilliant!

dokshina (approved):

PR #381, opened 2021-11-10
Fix CI

hackallcode:
Changes:
- use AWS instead of download.tarantool.io, because old TDG versions are now only available there;
- update version of TNT bundle, TDG;
- remove old CA certificate in Centos docker image;
- create caches of python dependencies;
- use pip2 instead of pip3 with python 2;
- specify Python dependencies for Python 2;
- fix CI matrix for Python 2;
- fix unit tests for Python 2;
- fix molecule tests for Python 2;
- update the last version of Ansible to 4.8.0 in CI;
- remove 2.10.0 Ansible in CI (it doesn't work with Molecule package on Python 2 and Python 3).

### DIFF
--- a/.github/helpers/count_molecule_matrix.py
+++ b/.github/helpers/count_molecule_matrix.py
@@ -22,23 +22,24 @@ NOT_DEFAULT_MOLECULE_COMMANDS = [
 
 DEFAULT_ANSIBLE_VERSION = '2.8.0'
 NOT_DEFAULT_ANSIBLE_VERSIONS = [
-    '2.9.0',
-    '2.10.0',
-    '4.1.0',
+    '2.9.27',
+    '4.8.0',
 ]
 
-DEFAULT_TARANTOOL_VERSION = '2.6'
+DEFAULT_PYTHON_VERSION = '2.7'
+
+DEFAULT_TARANTOOL_VERSION = '2.8'
 NOT_DEFAULT_TNT_VERSIONS = [
     '1.10'
 ]
 
-DEFAULT_SDK_VERSION = '2.6.2-124-g2c3b91439-r391'
+DEFAULT_SDK_VERSION = '2.8.2-0-gfc96d10f5-r429'
 NOT_DEFAULT_SDK_VERSIONS = []
 
 TDG_SCENARIOS = [
     'config_upload_tdg'
 ]
-DEFAULT_TDG_VERSION = '1.7.7-0-g76c31fca'
+DEFAULT_TDG_VERSION = '1.7.14-0-g2fcf1fc7'
 TDG_VERSIONS = {
     'config_upload_tdg': [
         # auth-token header; no common.app_version; no admin.upload_config_api;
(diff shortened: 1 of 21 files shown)


.github/helpers/count_molecule_matrix.py
```diff
@@ -22,23 +22,24 @@
 
 DEFAULT_ANSIBLE_VERSION = '2.8.0'
 NOT_DEFAULT_ANSIBLE_VERSIONS = [
-    '2.9.0',
-    '2.10.0',
-    '4.1.0',
+    '2.9.27',
+    '4.8.0',
```
dokshina, line 26: This change isn't described in the description, so I don't understand why these versions are tested now.

.github/helpers/count_molecule_matrix.py
```diff
@@ -58,39 +59,63 @@
...
-    return {
-        **get_matrix_base(molecule_scenario, ansible_version),
-        'tdg_version': tdg_version or DEFAULT_TDG_VERSION,
-        'molecule_command': get_molecule_command(molecule_command),
-    }
+def get_ce_params(
+    molecule_scenario=None, ansible_version=None, python_version=None,
+    tarantool_version=None, molecule_command=None,
+):
+    matrix = get_matrix_base(molecule_scenario, ansible_version, python_version)
```
dokshina, line 89: Why did you change the way these dictionaries are created?

.github/helpers/count_molecule_matrix.py
```diff
@@ -113,7 +138,10 @@ def main(event_name, repo_owner, review_state, ref):
             ce_matrix.append(get_ce_params(molecule_command=command))
 
         for version in NOT_DEFAULT_ANSIBLE_VERSIONS:
-            ce_matrix.append(get_ce_params(ansible_version=version))
+            python_version = None
+            if compare_versions(version, '2.10.0') >= 0:
```
dokshina: There are 2 "version" variables here, let's rename `version -> ansible_version`.

.github/helpers/count_molecule_matrix.py
```diff
@@ -125,13 +153,13 @@ def main(event_name, repo_owner, review_state, ref):
...
+    print('::set-output name=ce-tests-found::' + "true" if len(ce_matrix) > 0 else "false")
+    print('::set-output name=ee-tests-found::' + "true" if len(ee_matrix) > 0 else "false")
+    print('::set-output name=tdg-tests-found::' + "true" if len(tdg_matrix) > 0 else "false")
 
-    print(f'::set-output name=ce-matrix::{json.dumps({"include": ce_matrix})}')
-    print(f'::set-output name=ee-matrix::{json.dumps({"include": ee_matrix})}')
-    print(f'::set-output name=tdg-matrix::{json.dumps({"include": tdg_matrix})}')
+    print('::set-output name=ce-matrix::' + json.dumps({"include": ce_matrix}))
+    print('::set-output name=ee-matrix::' + json.dumps({"include": ee_matrix}))
+    print('::set-output name=tdg-matrix::' + json.dumps({"include": tdg_matrix}))
```
dokshina, line 162: This is because of using Python2?

.github/workflows/molecule-tests.yml
```diff
@@ -14,11 +14,12 @@ jobs:
       - name: Setup python
         uses: actions/setup-python@v2
         with:
-          python-version: '2.7'
+          python-version: '3.9'
```
dokshina, line 17: But here Python 3 is used for generating matrix.

.github/workflows/molecule-tests.yml
```diff
@@ -44,13 +45,28 @@ jobs:
...
         run: |
-          pip3 install --upgrade -r requirements.txt
-          pip3 uninstall -y ansible ansible-base ansible-core
-          pip3 install ansible~=${{ matrix.ansible_version }}
+          python -V
+          if [ "${{ matrix.python_version }}" = "2.7" ]; then
+              pip install virtualenv
+              virtualenv venv
+          else
+              python -m venv venv
```
dokshina, line 65: BTW, did you check GH Actions modules for python venv?

.github/workflows/unit-tests.yml
```diff
@@ -11,38 +11,56 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        tarantool-version: ["1.10"]
+        python_version: ["2.7", "3.9"]
```
dokshina, line 14: Why do you use `_` here, but `-` above?

molecule/rolling_update/converge.yml
```diff
@@ -13,7 +13,7 @@
     cartridge_scenario:
       - deliver_package
       - update_package
-    cartridge_package_path: ./packages/myapp-5.0.0-0-with-c-2.5.0.tar.gz
+    cartridge_package_path: ./packages/myapp-5.0.0-0-with-c-2.6.0.tar.gz
```
dokshina, line 16: I can be wrong, but the reason for testing special versions was the difference in the internal Cartridge API.

molecule/rolling_update/tests/test_dists_rotated.py
```diff
@@ -1,10 +1,24 @@
...
+
+from testinfra.modules.file import File
 
 import utils
 
 testinfra_hosts = utils.get_testinfra_hosts()
 
 
+def listdir(self):
+    out = self.run_test("ls -1 -q -- %s", self.path)
```
dokshina, line 12: Wasn't mentioned in description, I didn't understand.

dokshina (commented):
IMHO, there are enough changes to split this PR into 3 or 2.

.github/helpers/count_molecule_matrix.py
hackallcode: Because spread operator is not supported by Python 2

.github/helpers/count_molecule_matrix.py
hackallcode: Added:
```
- update the last version of Ansible to 4.8.0 in CI;
- remove 2.10.0 Ansible in CI (it doesn't work with Molecule package on Python 2 and Python 3);
```

.github/workflows/unit-tests.yml
hackallcode: Where? `_` is used everywhere

molecule/rolling_update/converge.yml
hackallcode: 2.4.0 doesn't work :)

More info here:
https://github.com/tarantool/sdk/issues/245

But it used only to update from the old version to the new version, so 2.5.0 + 2.6.0 is good too

molecule/rolling_update/tests/test_dists_rotated.py
hackallcode: This function doesn't exists in Python 2, so it's a copy-paste from Python 3 library

.github/workflows/molecule-tests.yml
hackallcode: What GH Actions modules do you mean? Actions? In any case, importing action is harder than this if

.github/workflows/molecule-tests.yml
hackallcode: Linter flake8 is run by Python 2 and Python 3. It checks this file too. So, this file should be compatible with both versions of Python.

.github/helpers/count_molecule_matrix.py
hackallcode: Yes

.github/helpers/count_molecule_matrix.py
hackallcode: Fixed

dokshina (approved):

PR #386, opened 2021-11-18
Add availability to configure log directory

### LINKED ISSUE #378: [3sp] Add `cartridge_log_dir` to the update_package step
`update_package` has the creation of `wal`, `run`, `conf`, `data`, `memtx` and even `vinyl` directories. But no logs... Sometimes we want to place logs onto a different mount, so It won't trash our `wal` dir. In that case we need to have a directory present.

For example, `cartridge_log_dir: /data/tarantool/logs/<app_name>/`

If I understand correctly all those directories are created on a condition if the variable is defined? Can we add logs then?

hackallcode:
Closes #378

Before this patch, it was not possible to specify a directory for storing instance logs.

Now you can specify the parent directory for logs, it will be automatically created,
and inside it there will be log files for all instances from current machine.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ Please update `ansible-galaxy install` command in
 README.md to use the newest tag with new release
 -->
 
+### Added
+
+- Add `cartridge_log_dir_parent` to configure directory of logs
+
 ### Fixed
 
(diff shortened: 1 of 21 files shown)


dokshina (approved):
Brilliant!

PR #389, opened 2021-11-29
Fix backup folder permissions

hackallcode:
Before the patch, it was impossible to backup data by a non-root user,
because the backup folder had permissions 660, which did not allow the directory to be read.

Now 750 permissions are used, so the folder can be read by the owner user or owner group.


dokshina (commented):
Please, update changelog =)

dokshina (approved):

no1seman:
> Before the patch, it was impossible to backup data by a non-root user, because the backup folder had permissions 660, which did not allow the directory to be read.
> 
> Now 755 permissions are used, so the folder can be read by the owner user or owner group.

750 not 755!

PR #393, opened 2022-02-17
Fix templates of systemd units for TGZ packages

### LINKED ISSUE #391: [1pt] Fix templates of systemd units for TGZ packages
Use:

```
User={{ cartridge_app_user }}
Group={{ cartridge_app_group }}
```

Instead of:

```
User=tarantool
Group=tarantool
```

in templates of systemd units for TGZ packages

hackallcode:
Closes #391

Before the patch, systemd files used the `tarantool` user and the `tarantool` group.

Not, this uses the value of the `cartridge_app_user` and `cartridge_app_group` variables.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ README.md to use the newest tag with new release
 - Optimize facts caching
 - Fixed the ability to roll back to the previous TGZ package
 - Fixed backup folder permissions
+- Fixed templates of systemd units for TGZ packages
 
 ## [1.11.0] - 2021-07-30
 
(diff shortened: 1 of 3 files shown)


olegrok (approved):
Почему без тестов?

hackallcode:
> Почему без тестов?

We do not have other users in the test environment than `tarantool`. So it will be difficult to test other users. However, all tests involving TGZ packages test that the `tarantool` user and `tarantool` group (they are by default) work correctly.

Additionally, I manually looked at the generated files, everything is correct.

PR #400, opened 2022-02-24
Add the ability to choose a control instance among the leaders

### LINKED ISSUE #396: Make possible to choose control instance only from masters
Some operations in cluster must be initiated from masters only. So need to add a flag to eval_on_control_instance step to be able to select any alive instance only from current masters in their replicasets.

hackallcode:
Closes #396

Before the patch, the control instance was chosen among all live instances.

Now added the ability to select a control instance only among leaders
using the `cartridge_force_leader_control_instance` flag

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -8,7 +8,7 @@ cartridge_not_save_cookie_in_app_config: false
 cartridge_remove_temporary_files: false
 cartridge_ignore_split_brain: false
 cartridge_paths_to_keep_on_cleanup: []
-cartridge_only_leader_controls: false
+cartridge_force_leader_control_instance: false
 
 # Role scenario configuration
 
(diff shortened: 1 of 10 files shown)


defaults/main.yml
```diff
@@ -8,6 +8,7 @@ cartridge_not_save_cookie_in_app_config: false
 cartridge_remove_temporary_files: false
 cartridge_ignore_split_brain: false
 cartridge_paths_to_keep_on_cleanup: []
+cartridge_only_leader_controls: false
```
dokshina: Maybe it's better to use "force" word here? Like cartridge_force_leader_control_instance.

doc/steps.md
```diff
@@ -236,7 +236,8 @@ This is instance that is:
   otherwise, any instance that should be joined during current play is chosen;
 * control instance should have minimal Cartridge version across all suitable
   instances (because Cartridge two-phase commit should be called by instance
-  that has lowest version).
+  that has lowest version);
+* control instance should be a leader if `cartridge_only_leader_controls` is set.
```
dokshina: should be -> is?

dokshina: What if there are no leader instances that have lowest Cartridge version? I think that in this case we should return an error.

dokshina (commented):
Great patch with tests and doc!

dokshina (approved):
Great!

PR #402, opened 2022-03-02
Add `cartridge_app_config_upload_http_timeout` variable

### LINKED ISSUE #401: Add availability to configure timeouts of apply config method
Seems that we should add a timeout for HTTP and patch clusterwide methods:
https://github.com/tarantool/ansible-cartridge/blob/e8771dc3a66dc55a5c0de24397c8cfc32d7cd3ff/library/cartridge_apply_app_config.py#L48
https://github.com/tarantool/ansible-cartridge/blob/e8771dc3a66dc55a5c0de24397c8cfc32d7cd3ff/library/cartridge_apply_app_config.py#L60

hackallcode:
Closes #401

Before the patch, in the `upload_app_config` step,
the default timeouts were used for the clusterwide config patch
and for uploading the config via HTTP.

Now, the `cartridge_app_config_upload_http_timeout` variable is used
as a config upload timeout for HTTP mode.

Also, the `upload_app_config` step currently uses `twophase_upload_config_timeout`,
`twophase_netbox_call_timeout` and `twophase_apply_config_timeout` variables.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ README.md to use the newest tag with new release
 - Fixed backup folder permissions
 - Handle empty values in `helpers.py`
 - Fixed templates of systemd units for TGZ packages
+- The `upload_config_timeout` variable is used in the `upload_app_config` step
 
 ## [1.11.0] - 2021-07-30
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@ README.md to use the newest tag with new release
 - Add `cartridge_log_dir_parent` to configure directory of logs
 - Add `cartridge_force_leader_control_instance` variable to choose a control
   instance among the leaders
+- Add `cartridge_app_config_upload_http_timeout` variable to configure timeout to
+  wait config upload in HTTP mode.
 
 ### Fixed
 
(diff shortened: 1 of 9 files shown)


CHANGELOG.md
```diff
@@ -24,6 +24,7 @@ README.md to use the newest tag with new release
 - Fixed backup folder permissions
 - Handle empty values in `helpers.py`
 - Fixed templates of systemd units for TGZ packages
+- The `upload_config_timeout` variable is used in the `upload_app_config` step
```
dokshina: Are sure in variable name?

doc/app_config.md
```diff
@@ -121,6 +121,9 @@ which is on the same machine as the control instance:
 cartridge_app_config_upload_url: 'http://10.0.0.102:8083/admin/config'
 ```
 
+If you have a large configuration or slow connection, you can increase
+the upload timeout with `cartridge_app_config_upload_http_timeout` variable.
+
```
dokshina, line 126: Maybe say about default timeout?

molecule/config_upload/converge.yml
```diff
@@ -27,6 +27,7 @@
         cartridge_app_config_path: './configs/config.yml'
         cartridge_app_config_upload_mode: null
         cartridge_app_config_upload_url: 'http://vm1:3301/admin/config'
+        cartridge_app_config_upload_http_timeout: 30
```
dokshina, line 30: Why do you explicitly set default value?

hackallcode: Due the bug in Ansible: cartridge_app_config_upload_http_timeout value will be share for neighboring calls

doc/app_config.md
hackallcode: fixed

CHANGELOG.md
hackallcode: fixed

dokshina (approved):

PR #409, opened 2022-05-04
Add step to backup stopped instances

### LINKED ISSUE #405: [4pt] Add step to backup not started instance
The current backup step can only back up running instances. In some cases this is useless.

hackallcode:
Closes #405

Now it's possible to backup stopped instance by `backup_instance_dirs` step.
This makes it possible to transfer instances from one host to another.

Also, now it's possible to restore instances from local backup by `restore` step with `cartridge_restore_backup_path_local` variable.


doc/steps.md
```diff
@@ -591,11 +592,32 @@ Input variables from config:
   (`.tarantool.cookie` will be kept independently of this variable);
   it's possible to use bash patterns, e.g. `*.control`.
 - `cartridge_restore_backup_path` - path to the instance backup archive on the remote machine;
+- `cartridge_restore_backup_path_local` - path to the instance backup archive on the local machine;
 - `cartridge_remote_backups_dir` - directory with backups on the remote;
 - `cartridge_force_restore` - flag indicates that conflicting files should be overwritten;
 - `cartridge_allow_alien_backup` - flag indicates that backup of instance with another name can be used;
 - `cartridge_skip_cleanup_on_restore` - flag indicates that cleanup before restoring should be skipped.
 
+### Step `backup_files`
```
dokshina: As for me, step name doesn't reflect the main idea that stopped instances are backed up.

library/cartridge_backup_instance.py
```diff
@@ -97,13 +101,31 @@ def backup_stop(control_console):
...
+    assert backups_dir is not None, 'Parameter "backups_dir" is required'
+    backup_archive_path = backup_pack(instance_id, backups_dir, custom_backup_files)
+
+    return helpers.ModuleRes(changed=True, fact={
+        'backup_archive_path': backup_archive_path,
+        'backup_files': custom_backup_files,
+    })
+
+
+def tnt_backup(params):
```
dokshina, line 118: All backups here are tnt. Maybe it's "backup_running_instance" and "backup_stopped_instance"?

library/cartridge_backup_instance.py
```diff
@@ -97,13 +101,31 @@ def backup_stop(control_console):
...
+    start_only = params.get('start_only', False)
+    stop_only = params.get('stop_only', False)
 
     assert not (start_only and stop_only), "impossible to use 'start_only' with 'stop_only'"
 
-    console_sock = params['console_sock']
+    console_sock = params.get('console_sock')
+    assert console_sock is not None, 'Parameter "console_sock" is required'
```
dokshina: What about smth like helpers.get_required_param(params, "name") instead of this copy-paste?

library/cartridge_get_instance_info.py
```diff
@@ -31,6 +31,12 @@ def get_instance_systemd_service(app_name, instance_name, stateboard=False):
     return '%s@%s' % (app_name, instance_name)
 
 
+def get_instance_systemd_service_template(app_name, stateboard=False):
```
dokshina, line 34: I don't understand why this function name contains "template"? It returns systemd service file name, isn't it?

molecule/move_instance/converge.yml
```diff
@@ -0,0 +1,109 @@
...
+
+    - name: 'Check fetched_backup_archive_path fact is empty'
+      assert:
+        fail_msg: 'Received bad fetched_backup_archive_path'
+        success_msg: 'Received fetched_backup_archive_path is OK'
+        that: fetched_backup_archive_path is none
+
+    # Backup files with fetching archive
+
+    - name: 'Call backup files'
```
dokshina, line 36: Stop instance and call backup

dokshina (commented):
Molecule test looks cool. This is very interesting scenario. I have some naming suggestions, like always. The main one is step name, I sure that it should reflect that it works with stopped instances.

library/cartridge_backup_instance.py
hackallcode: One type of backups is provided by Tarantool box.backup(), another type of backups is provided by our custom list of files. So, when you are using box.backup(), you should start instance. When you are using out custom list of files, you should stop instance

library/cartridge_get_instance_info.py
hackallcode: It returns path to systemd template file (without instance name, with %i variable). `get_instance_systemd_service` returns path to instance systemd service

dokshina (approved):

PR #415, opened 2022-06-20
Add disabled instances to `single_instances_for_each_machine` variable

hackallcode:
Before this patch, only enabled instances got into the `single_instances_for_each_machine` variable. This prevented updating instances that are disabled.

Now, if no enabled instance was found for a machine, a disabled instance is added to this list.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ README.md to use the newest tag with new release
 - Remove old app configurations before uploading a new one
 - Allow downgrading RPM and DEB packages
 - Ignore disabled instances when counting disabled instances
+- Add disabled instances to `single_instances_for_each_machine` variable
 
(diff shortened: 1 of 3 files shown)


dokshina (commented):
Is it Ok to disable updated instances? Is it possible to disable instance by it's own? What is the reason of having disable instances in single instances for machine list?

hackallcode:
> Is it Ok to disable updated instances?

This is the best way to update replicas as Cartridge will not send requests disabled, thus no errors will occur during the update.

> Is it possible to disable instance by it's own?

No, you should select one not disabled instance to disabled another from it

> What is the reason of having disable instances in single instances for machine list?

If I want to update only disabled replicas, I specify only them in play hosts. And if these replicas are not in the single instances for machine list, then no new package will be delivered to the machine

PR #426, opened 2023-07-18
Add `twophase_validate_config_timeout` variable __

### LINKED ISSUE #425: Can't set `validate_config_timeout`
On highloaded clusters we have problems with updating the clusterwide config with default values of timeouts.
We can increase some of them, for example: 
https://github.com/tarantool/ansible-cartridge/blob/bb85d618891461b7619567dbf088d10d939f629e/doc/variables.md?plain=1#L134-L139

But we can't set `validate_config_timeout`, which is bottleneck, since this timeout affects the "prepare" phase of applying CW-config.  This phase has a lot of implicit yields, since there are a lot of IO operations.

palage4a:
In some cases, when we work with highloaded cluster, which can't apply CW-config with default timeouts,
we have to increase all twophase-timeouts.

Before the patch, there were only three of four `cartridge.twophase` timeouts:
- `twophase_netbox_call_timeout` -> `netbox_call_timeout`;
- `twophase_upload_config_timeout` -> `updload_config_timeout`;
- `twophase_apply_config_timeout` -> `apply_config_timeout`;

So, we didn't have an ability to set the `validate_config_timeout` variable.

Closes #425


palage4a:
Seems that most of checks fall because of troubles with environment. Do we need fix it before merge?

hackallcode (approved):
The patch looks good, but for the merge it is necessary to fix CI in a separate PR

